PR #4, opened 2023-10-25
feat: added coverage viz summary

saq7:
# Pull Request

## JIRA Ticket

[CZID-8259](https://czi-tech.atlassian.net/browse/CZID-8259)

## Description

Added Coverage viz summary entity. 

Also added a new http method that does not add a prefix to the request.

## Notes

## Tests

[CZID-8259]: https://czi-tech.atlassian.net/browse/CZID-8259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


utils/httpUtils.ts
```diff
@@ -27,6 +27,23 @@ export const get = async (url: string, args: any, context: any) => {
...
+        const response = await fetch(baseURL + url, {
+            method: 'GET',
+            headers: {
+              'Cookie': context.request.headers.get("cookie"),
+              'Content-Type': 'application/json',
+            },
+          });
+        return await response.json();
+    } catch (e) {
+        console.log(e);
```
jfoo1984, line 42: (nit) could we use `console.error`?  I don't know if we have linting rules setup, but typically we'd avoid having `console.log` in prod

utils/httpUtils.ts
```diff
@@ -27,6 +27,23 @@ export const get = async (url: string, args: any, context: any) => {
     }
 };
 
+export const simpleGet = async (url: string, context: any) => {
```
jfoo1984, line 30: Could we just reuse the `get` function above?  It looks like as long as you don't pass an `args` object with the `snapshotLinkId` key, it should work the same as the `simpleGet` method.

.meshrc.yaml
```diff
@@ -57,6 +57,12 @@ sources:
             method: GET
             responseSchema: ./json-schemas/mngsWorkflowResults.json
             responseTypeName: MngsWorkflowResults
+          - type: Query
+            field: CoverageVizSummary
+            path: /samples/{args.sampleId}/coverage_viz_summary
+            method: GET
+            responseSchema: ./json-schemas/coverageVizSummary.json
+            responseTypeName: CoverageVizSummary
```
jfoo1984, line 65: I think coverage viz summary may also be available at the endpoint `/pub/[snapshotShareId]`.  It [appears to call the samples controller](https://github.com/chanzuckerberg/czid-web-private/blob/a4dd2a773881ee9f2b5a5707225f15280e052903/app/controllers/snapshot_samples_controller.rb#L149-L152). but should we add a config to federate that as well?

saq7: ah wait, that ticket was created for the entity it looks like. Will add this endpoint

saq7: Looking into this a bit further, I am not sure why that endpoint is part of the ticket. It renders a template

saq7: The endpoint at `GET /pub/:share_id/samples/:id/coverage_viz_summary` does not need to be federated since the implementation of it calls the same method (via inheritance) as the one federated here

saq7: Federated public endpoint

resolvers.ts
```diff
@@ -141,6 +236,32 @@ export const resolvers: Resolvers = {
...
+          ]) {
+            return_obj.push({
+              pipeline_id: key,
+              ...accension,
+            });
+          }
+        }
+        return return_obj;
+      } catch (error) {
+        console.log(error);
```
irene-han, line 261: Should this console log be merged?

saq7: removed

resolvers.ts
```diff
@@ -141,6 +236,31 @@ export const resolvers: Resolvers = {
...
+          ]) {
+            return_obj.push({
+              pipeline_id: key,
+              ...accension,
+            });
+          }
+        }
+        return return_obj;
+      } catch (error) {
+        return [];
```
jfoo1984, line 261: I know it was asked for the `console.log` to be removed, but should we return the error still?  I'm not sure if it's different with relay, but typically graphQL responses have an errors field that is an array right?

jfoo1984: I looked into error handling a bit, and here's some recommended options I found

- graphQL mesh [documents error handling in custom resolvers](https://the-guild.dev/graphql/mesh/docs/guides/error-masking#custom-resolvers-error-handling)
- Relay [recommends creating types for error information as part of the schema](https://relay.dev/docs/guided-tour/rendering/error-states/#accessing-errors-in-graphql-responses)
- The graphQL spec includes a [top level `errors` entry in the response that includes a list of errors](http://spec.graphql.org/draft/#sec-Errors), as well as some other information depending on the error.

I didn't try this out, but it's possible the graphQL mesh approach ends up generating an error field that matches the graphQL spec.  This might be worth getting some other inputs on from the front end team.

saq7: good call, let's just remove the try/catch and make it consistent with the other resolvers.

resolvers.ts
```diff
@@ -141,6 +236,25 @@ export const resolvers: Resolvers = {
       });
       return annotations;
     },
+    CoverageVizSummary: async (root, args, context, info) => {
+      // should be fetched using pipeline run id instead of sample id
+      // from the new backend
```
jfoo1984, line 241: Is this not possible to do in the current Rails endpoint (without making modifications to Rails)?  Would it be worth adding `pipeline_run_id` argument in as a placeholder (though it'd be unused for now)?

resolvers.ts
jfoo1984: Just to make sure, the error here would've been some TS error, not an error coming back from the API or something like that?

jfoo1984 (approved):
Just had a couple of questions

resolvers.ts
saq7: We don't know the exact structure of the new endpoint yet. I think it's better to integrate wholesale with the new backend as opposed to doing it piecemeal

PR #5, opened 2023-10-27
feat(ci): promote images from staging when deploying to prod

jfoo1984:
# Pull Request
[CZID-8804]

## Description
Implement prod deployment by promoting images from staging to prod.  Locks our deploy action to use the version containing chanzuckerberg/github-actions/pull/228, which enables a new argument needed for this feature.

[CZID-8804]: https://czi-tech.atlassian.net/browse/CZID-8804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Notes

* Set `IMAGE_SOURCE_ROLE_ARN` as Github secret in production environment
* `IMAGE_SOURCE_ENV` and `IMAGE_SOURCE_STACK` were previously set.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -38,7 +38,7 @@ jobs:
         run: |
           HAPPY_OIDC_ID_TOKEN=${{env.id_token}} happy config set API_URL ${{ vars.API_URL }} --env prod --aws-profile ""
       - name: Deploy to production env
-        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@main
+        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.26.0
         with:
           tfe-token: ${{ secrets.TFE_TOKEN }}
           env: ${{ vars.HAPPY_ENV }}
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -42,6 +42,8 @@ jobs:
         with:
           tfe-token: ${{ secrets.TFE_TOKEN }}
           env: ${{ vars.HAPPY_ENV }}
-          create-tag: true
+          image-source-env: ${{ vars.IMAGE_SOURCE_ENV }}
+          image-source-stack: ${{ secrets.IMAGE_SOURCE_STACK }}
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-prod.yml
```diff
@@ -38,10 +38,12 @@ jobs:
         run: |
           HAPPY_OIDC_ID_TOKEN=${{env.id_token}} happy config set API_URL ${{ vars.API_URL }} --env prod --aws-profile ""
       - name: Deploy to production env
-        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@main
+        uses: chanzuckerberg/github-actions/.github/actions/deploy-happy-stack@v1.26.0
         with:
           tfe-token: ${{ secrets.TFE_TOKEN }}
           env: ${{ vars.HAPPY_ENV }}
-          create-tag: true
```
vincent-czi, line 45: I see we're dropping `create-tag: true` entirely with this PR. I don't really know the specifics of `deploy-happy-stack`, but is this something we'd want to keep with what this PR is adding? It feels like if we wanted a tag created before, we would still want one now, but the new args don't seem to have anything to do with tagging one way or the other. Just wanted to make sure this was purposeful and not an accidentally cut out line!

vincent-czi (approved):
Seems reasonable! Long term, it would be nice to eventually unpin and get back to being on `@main` rather than staying pinned to a specific version of the deploy-happy-stack action, just in case there's eventually a security patch or something we might forget about.

jfoo1984:
> Long term, it would be nice to eventually unpin and get back to being on @main rather than staying pinned to a specific version of the deploy-happy-stack action, just in case there's eventually a security patch or something we might forget about.

I believe the happy team actually recommended that I pin to a version, because they make make breaking changes, which would break our deploy pipelines if we automatically use the latest version

.github/workflows/deploy-prod.yml
jfoo1984: `create-tag` defaults to false

https://github.com/chanzuckerberg/github-actions/blob/0b740d81436af68cb9af1011cab89199823c0161/.github/actions/deploy-happy-stack/action.yml#L11-L14

To use the image promotion, `create-tag` does need to be set to false, so I've just dropped it here and relied on the default.

PR #9, opened 2023-10-30
feat: added kickoff_workflow

saq7:
# Pull Request

## JIRA Ticket

[CZID-8261](https://czi-tech.atlassian.net/browse/CZID-8261)

## Description

Federated Kickoff Workflow endpoint

## Notes

## Tests

[CZID-8261]: https://czi-tech.atlassian.net/browse/CZID-8261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.meshrc.yaml
+++ b/.meshrc.yaml
@@ -127,6 +127,15 @@ sources:
             responseTypeName: UpdateSampleNotes
             headers:
               Cookie: "{context.headers['cookie']}"
+          - type: Mutation
+            field: KickoffWorkflow
+            path: /samples/{args.sampleId}/kickoff_workflow
+            method: POST
+            requestSample: {"inputs_json[start_from_mngs]": "true", "workflow": "amr", "authenticityToken": "token"}
+            responseSchema: ./json-schemas/kickoffWorkflow.json
+            responseTypeName: UpdateSampleNotes
+            headers:
+              Cookie: "{context.headers['cookie']}"
       transforms:
         - namingConvention:
             typeNames: camelCase
--- a/.meshrc.yaml
+++ b/.meshrc.yaml
@@ -131,7 +131,7 @@ sources:
             field: KickoffWorkflow
             path: /samples/{args.sampleId}/kickoff_workflow
             method: POST
-            requestSample: {"inputs_json[start_from_mngs]": "true", "workflow": "amr", "authenticityToken": "token"}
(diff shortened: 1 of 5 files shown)


.meshrc.yaml
```diff
@@ -141,6 +141,15 @@ sources:
             responseTypeName: UpdateSampleNotes
             headers:
               Cookie: "{context.headers['cookie']}"
+          - type: Mutation
+            field: KickoffWorkflow
+            path: /samples/{args.sampleId}/kickoff_workflow
+            method: POST
+            requestSample: {"inputs_json": {}, "workflow": "amr", "authenticityToken": "token"}
+            responseSchema: ./json-schemas/kickoffWorkflow.json
+            responseTypeName: UpdateSampleNotes
```
jfoo1984, line 150: is this the correct value?  looks like it might be copy-pasta from above.

.meshrc.yaml
```diff
@@ -141,6 +141,15 @@ sources:
             responseTypeName: UpdateSampleNotes
             headers:
               Cookie: "{context.headers['cookie']}"
+          - type: Mutation
+            field: KickoffWorkflow
```
jfoo1984, line 145: This endpoint is actually called for kick off WGS and AMR (see linked code below).  Could we create two separate mutations and name them specifically, so maybe `KickoffWGSWorkflow` and `KickoffAMRWorkflow`?  Also, could we add a type for the `inputs_json` field, since they are different for each one?

Technically, we only need WGS for now, but if it's not too much lift, it'd probably make sense to go ahead and federate AMR as well since they're pretty similar.  That way we don't have to come back to this later on and built context again.

https://github.com/chanzuckerberg/czid-web-private/blob/4d6435761c594bb4214b2a5b30622e1cfc028a9d/app/assets/src/api/index.ts#L716-L744

saq7: We don't need, or generally add types, for input arguments. That gets picked up automatically.

jfoo1984: It does get picked up automatically, but in this case, since the inputs are different depending on pipeline, it would be helpful to type them.

Also, nothing stops us from setting up two separate mutation fields, as I mentioned above, that call the same REST endpoint right?  When we convert to graphQL in React, we would be able to call a workflow specific mutation, rather than a generic kickoff workflow.

jfoo1984 (approved):

PR #16, opened 2023-11-08
chore: swap sampleId type to ID to support QualityControlQuery

irene-han:
# Pull Request

## JIRA Ticket

[CZID-8652]

## Description

Change graphql scheme to match changes in this [PR](https://github.com/chanzuckerberg/czid-web-private/pull/4052/)

## Notes
This is a non-breaking change. The federated `sampleReadsStats` query is not currently being used in the web-app.

Note that this change must be merged before the associated PR on the web-app side can be merged.

## Tests
Tested locally

[CZID-8652]: https://czi-tech.atlassian.net/browse/CZID-8652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sources/czid-schema.graphql
+++ b/sources/czid-schema.graphql
@@ -208,7 +208,7 @@ type Query {
   pathogenList(version: String): PathogenList!
   project(id: Int!): Project!
   sample(sampleId: Int!): Sample!
-  sampleReadsStats(sampleIds: [Int!]!): SampleReadsStatsList!
+  sampleReadsStats(sampleIds: [String!]!): SampleReadsStatsList!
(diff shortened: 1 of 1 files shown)


sources/czid-schema.graphql
```diff
@@ -208,7 +208,7 @@ type Query {
   pathogenList(version: String): PathogenList!
   project(id: Int!): Project!
   sample(sampleId: Int!): Sample!
-  sampleReadsStats(sampleIds: [Int!]!): SampleReadsStatsList!
+  sampleReadsStats(sampleIds: [String!]!): SampleReadsStatsList!
```
jfoo1984, line 211: is it valid to make this [ID!]?  or do we have to do this since it's being returned upstream as an array of strings?

jfoo1984 (approved):
was the schema update generated with `aws-oidc exec --profile idseq-dev-poweruser -- docker compose exec web bin/rails graphql:schema:dump`, or at least match the output of that?  I believe this is how we originally got this file.

jfoo1984:
I think this actually should've been marked as a breaking change, to force a release.  It happens not to break anything for us since the query isn't being actively used.

rainandbare:
It is actually being used Jerry! The change breaks the Quality Control query in staging.

jfoo1984:
Did we already merge the switch of this call to Relay in CZ ID?  The Apollo queries should be going straight to Rails, and not through graphQL federation.

PR #21, opened 2023-11-13
feat!: swap sampleId type to ID to support QualityControlQuery

jfoo1984:
Reverts chanzuckerberg/czid-graphql-federation-server#20

Re-releasing #16 as a breaking change.  As mentioned in #16, this will not break anything in CZ ID, because the graphQL mesh federated query is not currently in use, but changing types in a graphQL schema should generally be considered a breaking change, and we do want this to be explicitly included in a release.

### DIFF
--- a/sources/czid-schema.graphql
+++ b/sources/czid-schema.graphql
@@ -208,7 +208,7 @@ type Query {
   pathogenList(version: String): PathogenList!
   project(id: Int!): Project!
   sample(sampleId: Int!): Sample!
-  sampleReadsStats(sampleIds: [Int!]!): SampleReadsStatsList!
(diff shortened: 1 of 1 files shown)


rainandbare (commented):
Were you able to satisfy this questions Jerry? https://github.com/chanzuckerberg/czid-graphql-federation-server/pull/16#pullrequestreview-1723638694

rainandbare:
> this will not break anything in CZ ID, because the graphQL mesh federated query is not currently in use

This type is in use I believe and it will break the Quality Control page. Would you do anything different in that case?

rainandbare (approved):

PR #22, opened 2023-11-14
fix: updated all json schemas 

saq7:
# Pull Request

## JIRA Ticket

[CZID-8899]

## Description
The type names being used in the json-schema files are not correct. it uses python’s names (int, str, etc) where it should use the ones the defined in json schema spec.

For reference, these are the types defined in the JSON schema spec: https://json-schema.org/understanding-json-schema/reference/type

## Tests

[CZID-8899]: https://czi-tech.atlassian.net/browse/CZID-8899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


json-schemas/amrDeprecatedResults.json
```diff
@@ -1,50 +1,47 @@
...
+            "type": "integer"
+        },
+        "rpm": {
+            "type": "number"
+        },
+        "dpm": {
+            "type": "number"
         }
     }
 }
```
jfoo1984, line 47: (nit) Can we add newlines at the end of all files for consistency?  It appears some have it, so I assume it is allowed for JSON files

jfoo1984 (approved):
looks good to me, one small comment.  

Also, could we fill out the testing section of the PR description, even if to say there's no additional testing required?

jfoo1984:
@rainandbare once merged, we should be able to generate a new graphQL schema for Relay that will fix the field types that were showing up as JSON.

PR #25, opened 2023-11-24
feat: federated /workflow_runs/{args.workflowRunId}/zip_link.json

saq7:
# Pull Request

## JIRA Ticket

[CZID-8507](https://czi-tech.atlassian.net/browse/CZID-8507)

## Description

This is a request, federates the `/zip_link.json` endpoint.

It converts the header based location response to a regular body response.

## Notes

## Tests

Testing for this and point can be conducted using the graphiQL interface.

[CZID-8507]: https://czi-tech.atlassian.net/browse/CZID-8507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


resolvers.ts
```diff
@@ -263,6 +264,19 @@ export const resolvers: Resolvers = {
...
     },
+    ZipLink: async (root, args, context, info) => {
+      const res = await getFullResponse(`/workflow_runs/${args.workflowRunId}/zip_link.json`, args, context)
+      if (res.status !== 302) {
+        return {
+          url: null,
+          error: res.statusText,
+        };
+      }
+      const url = res.headers.get("location");
```
rainandbare, line 275: Why is this not `res.url`?

resolvers.ts
```diff
@@ -263,6 +264,19 @@ export const resolvers: Resolvers = {
...
       return return_obj;
     },
+    ZipLink: async (root, args, context, info) => {
+      const res = await getFullResponse(`/workflow_runs/${args.workflowRunId}/zip_link.json`, args, context)
+      if (res.status !== 302) {
+        return {
+          url: null,
+          error: res.statusText,
+        };
+      }
```
rainandbare, line 274: So even if the response is 200 you want to deliver `null` as the url field?

rainandbare (commented):
I tried the query on my GraphiQL interface and I wasn't able to make it work as is. I might be missing some context though.

resolvers.ts
saq7: Yes, in this case. A 200 will actually represent an error given how the rails endpoint is set up. It returns a 302 with the url in the header.

resolvers.ts
saq7: We don't need the url of the request in this case. We need the presigned s3 url that was returned in the header.

rainandbare: Ah Interesting. As it stands I am getting this response from a valid workflowRunId: 
`{
  "data": {
    "ZipLink": {
      "url": null,
      "error": "OK"
    }
  }
}`

I see in the `res` object from `getFullResponse` the appropriate s3 url and can access it by console logging or outputting `res.url`. 

What would be the best way to adjust the code to make a successful call?

jfoo1984:
Could we fill out the PR description to provide context on what this PR is for and how we will test it?

rainandbare (approved):

PR #27, opened 2023-11-29
fix: Updated schema for kick off workflow

saq7:
# Pull Request

## JIRA Ticket

[CZID-XXXX]

## Description

This PR fixes a bug with the json-schema where it was using an object field as the top level type instead of arrays.

Fixes another issue where mesh converts a string to date.

## Notes

## Tests


rainandbare (approved):
Worked for me in our code pairing - can't see any breaking changes.

.meshrc.yaml
```diff
@@ -163,7 +163,7 @@ sources:
                     "accession_name": "Hubei mosquito virus 2 strain mosZJ35453 segment 1 hypothetical protein 1 and hypothetical protein 2 genes, complete cds",
```
jfoo1984, line 163: this may be more of a general question about mutations, do we have to provide a request sample, or can we provide a json schema for the request?  If we can, it seems like it would be more consistent to do that,

PR #43, opened 2023-12-14
feat: added save_metadata_v2 endpoint

rainandbare:
# Pull Request

## JIRA Ticket

[CZID-2831](https://czi-tech.atlassian.net/browse/CZID-2831)

## Description

Added save_metadata_v2 endpoint 🐥🎉

## Notes

This is my first attempt at adding an endpoint. I wonder if we want to return something different than:  {"status": "success",
      "message": "Saved successfully"}. Is there some thinking on API design where we have touched on this?

## Tests

Works with this query: 
```
mutation UpdateMetadataMutation{
  UpdateMetadata(
    sampleId: [insert your sampleId],
    input: {
      field: "nucleotide_type",
      value: "DNA",
      authenticityToken: [fill in token]
    }
  ){
    status
    message
  }
}
```

[CZID-2831]: https://czi-tech.atlassian.net/browse/CZID-2831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


saq7 (approved):
Thanks for adding this. It looks great. 

Re your comment around the return data - for mutations and POST/PATCH/PUT in general, the only real information you need to send back is that it is successful. 

If it is not successful, then some guidance regarding what to do to make it successful. 

If there is some data that the frontend needs as a result of this mutation, that it doesn't already have (such as ID of the record created, or updated_at timestamp), that can be included.

PR #44, opened 2023-12-14
feat: Feature Flag based routing

saq7:
# Pull Request

## JIRA Ticket

[CZID-8731](https://czi-tech.atlassian.net/browse/CZID-8731)

## Description

## Notes

## Tests


Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN npm ci --verbose --no-optional && npm cache clean --force
 
 COPY . .
 
+RUN npx mesh build
```
jfoo1984, line 11: this is currently being run in the `entrypoint.sh`: https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/9af53fe5d9b8c0f88659014e054c430f708e1cfd/entrypoint.sh#L2

[`npm run build` script defined in package.json](https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/9af53fe5d9b8c0f88659014e054c430f708e1cfd/package.json#L7).  

Is there any reason we want to run build a separate time here?

I think we do want a single command to run build and start (which is why they're in a script together), to make sure the server is built with the latest schema.

docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
     restart: always
     environment:
       API_URL: http://web:3001
+      NEXTGEN_API_URL: http://web:8008
```
jfoo1984, line 13: I don't know if we're looking to set this up in deployed environments at this time, but we'll eventually need to add vars in each github env in this, and configure our github actions to pick up the variables and set them with happy config https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/9af53fe5d9b8c0f88659014e054c430f708e1cfd/.github/workflows/deploy-sandbox.yml#L34-L39

.meshrc.yaml
```diff
@@ -206,9 +220,9 @@ additionalTypeDefs: |
   }
 additionalResolvers:
   - ./resolvers.ts
-plugins:
-  - csrfPrevention:
-      requestHeaders: ['x-graphql-yoga-csrf']
+# plugins:
+#   - csrfPrevention:
+#       requestHeaders: ['x-graphql-yoga-csrf']
```
jfoo1984, line 225: I know this PR is WIP, but we do need to enable this before we can merge the PR.

jfoo1984:
I believe the general approach that we decided on was that the feature flag will be passed as an argument from the client (React app), since it already retrieves the feature flags as part of loading pages.  Are there specific use cases we had in mind where graphQL mesh needs to fetch the feature flags from the CZ ID Rails app?

Dockerfile
saq7: This can be removed I guess. I added it there because I was running into an issue where it complained that the assets weren't already built. This happened after I had deleted the build folder

docker-compose.yml
saq7: Yeah, it makes sense to hold off deploying it but I added it here so that it is available for local development.

saq7:
@jfoo1984 I don't think we have a specific usecase where the feature flag needs to be fetched by the federation server. Though I added it here to make it easier to develop and use as needed until the end-to-end flow wrt to the fetching of the feature flag is complete. We can remove this down the road.

utils/httpUtils.ts
```diff
@@ -1,69 +1,129 @@
...
+        Cookie: context.request.headers.get("cookie"),
+        "Content-Type": "application/json",
+      },
     });
-}
+    return await response.json();
+  } catch (e) {
+    return Promise.reject(e.response);
+  }
+};
```
jfoo1984, line 107: This function fetches feature flags from Rails right?  Do we need this?  I thought that we are expecting all feature flags to be passed from the front end.

.meshrc.yaml
```diff
@@ -13,6 +13,20 @@ sources:
         source: ./sources/czid-schema.graphql
         operationHeaders:
           Cookie: "{context.headers['cookie']}"
+  - name: NextGenEntities
+    handler:
+      graphql:
+        endpoint: ${NEXTGEN_API_URL}/graphql
+        source: ./sources/nextgen-entities-schema.graphql
+        operationHeaders:
+          Cookie: "{context.headers['cookie']}"
```
jfoo1984, line 22: This is already configured above, as `CZIDENTITIES`.  i don't know if they are exactly the same, but we should figure out which one is most up to date.  I think @rzlim08 added the one above.

.meshrc.yaml
jfoo1984: bump on this, we need to uncomment this and make sure things work.

Dockerfile
jfoo1984: bump, can this be removed without issue?

docker-compose.yml
jfoo1984: Yea, we do have to have this for local dev regardless, but my comment pertained to deployed environments.  for dev and sandbox, #96 added these env vars as part of the TF config

utils/httpUtils.ts
```diff
@@ -40,15 +60,35 @@ export const get = async (url: string, args: any, context: any) => {
...
+        method: "POST",
+        headers: {
+          Cookie: context.request.headers.get("cookie"),
+          "Content-Type": "application/json",
+          "X-CSRF-Token": args?.input?.authenticityToken,
+          Authorization: `Bearer ${czidServicesToken}`,
+        },
+        body: JSON.stringify(query),
+      });
+      return response;
```
jfoo1984, line 78: would it be viable to put this code in the `callGraphqlSource` util function instead, and call that util function here?  This util file is getting fairly long and complex, so it'd be nice to extract the chunks that make sense.

https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/010967bd3011dd1544a6bad984f4e64e2f3f0b08/utils/callGraphqlSource.ts#L3

utils/httpUtils.ts
```diff
@@ -91,3 +147,40 @@ export const notFound = (message: string) => {
...
+
+export const isNextGenEnabled = async (context) => {
+  let readFromNextGen = getFeatureFlagsFromRequest(context);
+  if (readFromNextGen !== null) {
+    // if the header is set, return the value
+    return readFromNextGen;
+  }
+  try {
+    const featureFlags = await getFeatureFlags(context);
+    const combinedFeatureFlags = featureFlags["launched_feature_list"].concat(
```
bchu1, line 178: Question - why are there 2 places where the `next_gen` flag could be?

utils/httpUtils.ts
```diff
@@ -22,16 +23,35 @@ export const formatUrlParams = (params: { [s: string]: unknown }) => {
...
-      method: "GET",
-      headers: {
-        Cookie: context.request.headers.get("cookie"),
-        "Content-Type": "application/json",
-      },
-    });
-    return await response.json();
+    let baseURL, urlPrefix;
+    const nextGenEnabled = await isNextGenEnabled(context);
+    if (!nextGenEnabled) {
```
bchu1, line 28: noticed that the other ones do `nextGenEnabled` first, might be nice to do that here too

utils/httpUtils.ts
```diff
@@ -40,15 +60,35 @@ export const get = async (url: string, args: any, context: any) => {
...
-        "Content-Type": "application/json",
-      },
-    });
-    return response;
+    const nextGenEnabled = await isNextGenEnabled(context);
+    if (nextGenEnabled) {
+      // next gen details
+      const czidServicesToken = await getEnrichedToken(context);
+      const query = context.params.query;
+      const response = await fetch(process.env.API_URL + url, {
```
jfoo1984, line 68: This is not the correct URL for nextGen, because this is concatenating the Rails base URL with the endpoint passed in, right?  Since we don't actually check the the flag until we get to this method, the resolvers won't actually know which URL to call.  Though for calling nextGen, since it's graphQL, I don't think we have to. worry about a different URL, it's always going to be `/graphql`.

jfoo1984:
Can we add unit tests for all of the new code here as well?

utils/httpUtils.ts
jfoo1984: I thought we are planning to only get the feature flag via the headers, and not querying them from Rails.  Do we need that functionality at all?

jfoo1984: I just saw @saq7 's previous comment about this: https://github.com/chanzuckerberg/czid-graphql-federation-server/pull/44#issuecomment-1889595336

I'm skeptical we should add this unless we had a particular use in mind.  We can save off the code for fetching from Rails in a branch or wiki if we'd like, but I'd recommend we just stick to the header pattern and require clients to make requests that way, until we encounter a use case where that won't work.

rainandbare (commented):
Have we tested to make sure that all the current rails calls are working as expected?

.meshrc.yaml
saq7: Thanks for the call out. I updated the details here in the two sources from Ryan's and removed that one. Will update the schemas as well

utils/httpUtils.ts
saq7: That is how the rails endpoint is set up. `launched_feature_list` is a universal list for all users and `allowed_feature_list` is user specific

utils/httpUtils.ts
saq7: sorry, oversight. Fixed.

utils/httpUtils.ts
```diff
@@ -91,3 +147,40 @@ export const notFound = (message: string) => {
...
+  } catch (e) {
+    return Promise.reject(e.response);
+  }
+};
+
+export const getFeatureFlagsFromRequest = (context) => {
+  return context.request.headers.get("readFromNextGen");
+};
+
+export const isNextGenEnabled = async (context) => {
```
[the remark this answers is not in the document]
saq7: we can rename it. But what other functionality are you envisioning?

utils/httpUtils.ts
saq7: I commented out that code for the merge. I suggest we leave the comments in place so that they can be enabled for local development until we are in the flow of building using this pattern

saq7:
@rainandbare yes the current rails calls should work as expected. Let me know if you run into any issues

.meshrc.yaml
```diff
@@ -1,18 +1,25 @@
...
+      graphql:
+        endpoint: ${NEXTGEN_ENTITIES_API_URL}/graphql
+        source: ./sources/nextgen-entities-schema.graphql
+        operationHeaders:
+          Cookie: "{context.headers['authorization']}"
+  - name: NextGenWorkflows
+    handler:
+      graphql:
+        endpoint: ${NEXTGEN_WORKFLOW_API_URL}/graphql
+        source: ./sources/nextgen-entities-schema.graphql
```
rzlim08, line 20: This should be the `workflows` schema

docker-compose.yml
rzlim08, line 13: I don't think `NEXTGEN_API_URL` is necessary anymore

.meshrc.yaml
```diff
@@ -1,18 +1,25 @@
 sources:
-  - name: CZIDENTITIES
-    handler:
-      graphql:
-        endpoint: ${NEXTGEN_ENTITIES_URL}/graphql
-        source: ./sources/entities-schema.graphql
```
rzlim08, line 6: Could you delete this schema if it's not being used anymore?

utils/httpUtils.ts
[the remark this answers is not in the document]
jfoo1984: probably if we end up needing feature flags around writing to nextGen that are separate from the read feature flags

rainandbare (approved):

utils/httpUtils.ts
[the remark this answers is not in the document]
rainandbare: Let's merge this in for now and make those variable name changes with any tests that need to be written.

jfoo1984:
To capture context here, I think we should test this PR integrated on sandbox and make sure it works before we merge it.  unfortunately, this was already merged, so it's now being reverted

utils/httpUtils.ts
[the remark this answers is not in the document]
jfoo1984: stated in a comment on the whole PR, but we don't need to rush to merge this.  we can (and should IMO) deploy this to sandbox and make sure the integration works, and clean things up before merging

PR #51, opened 2023-12-18
chore: Minor schema updates for deletion flow

bchu1:
Removes the _ from `_workflow`.

Corrects the `error` type to string.

Sets `validIds` and `invalidSampleNames` to required.


.meshrc.yaml
```diff
@@ -137,7 +137,7 @@ sources:
             requestSample:
               {
                 "ids": [1],
-                "_workflow": "short-read-mngs",
```
jfoo1984, line 140: It looks like this was added in #9, so perhaps @saq7 might have some context.  I believe we might've used the leading `_` convention to indicate fields that are not mapped anywhere in the new data model (so that we can tell which fields may need to be updated when we switch over to it), but I'm not sure if that is the reason here, since it is a mutation.

jfoo1984 (approved):
this looks good to me, one comment about the `_`

.meshrc.yaml
bchu1: I see, yeah I looked around at other fields and I didn't see any underscores, so it seemed like this was the only one.

I updated the schema json to mark the 2 arrays as required, just FYI

PR #38, opened 2023-12-08
feat: restrict CORS origin for graphQL mesh

jfoo1984:
# Pull Request

## JIRA Ticket

[CZID-8925]

## Description

Restrict the CORS Origins for graphQL mesh to address Armor Code security alert.  See JIRA ticket for details.

## Notes
* Must add new `ALLOWED_CORS_ORIGINS` env var for each environment
  * [x] sandbox: `https://sandbox.czid.org`
  * [x] staging: `https://staging.czid.org` 
  * [x] production: `https://czid.org`

## Tests

* Manually test that graphQL queries in playground and webapp work
  * [x] tested locally
  * [x] tested sandbox

[CZID-8925]: https://czi-tech.atlassian.net/browse/CZID-8925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #52, opened 2023-12-18
feat(ci): Set ALLOWED_CORS_ORIGINS app config

jfoo1984:
# Pull Request

## JIRA Ticket

[CZID-8925]

## Description
#38 included a fix for arbitary CORS origins, but did not set the env var needed to enable the feature in graphQL mesh.  This PR sets that env var via `happy config`, and also adds a reusable action for setting env vars.

## Tests

Manually test that graphQL queries in playground and webapp work

- [ ] tested sandbox

[CZID-8925]: https://czi-tech.atlassian.net/browse/CZID-8925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jfoo1984:
Now that the allow origins are actually being set in an env var,  I've been trying to override the origin using the browser dev tools to see if I can generate a request that fails.  I haven't been able to so far, though I'm not sure if that is a valid way to override the origin.  

From earlier work on localhost, I do know that the origin policy does block.  The playground on `localhost` would not work when only `web` was specified as an origin (`web` is the hostname of the webapp on the docker compose network).

jacoblerner-czi (approved):
Wasn't able to confirm that this actually fixed the CORS issue, but we should be able to get more definitive answers after we deploy and scan. Regardless, it's a good idea to merge this change so that we are correctly setting headers from the config.

PR #66, opened 2024-01-12
Added sample response from GET /projects.json

saq7:
# Pull Request

## JIRA Ticket

[CZID-XXXX]

## Description

## Notes

## Tests


jfoo1984:
I know we like to include the response as is, but this file is rather large.  Could we just keep maybe a few of the projects around and manually delete the rest?  That should be enough to get the idea and generate JSON schema from.

PR #61, opened 2024-01-06
deps: update dependencies

jfoo1984:
# Pull Request

## JIRA Ticket

N/A

## Description

Update `graphql-mesh` dependecies to latest

## Tests
* deployed to sandbox, am able to make a few basic queries

PR #67, opened 2024-01-13
chore: Add jest unit testing

jfoo1984:
# Pull Request

## JIRA Ticket

[CZID-9015]

## Description

Add unit tests based on [graphQL mesh examples](https://github.com/ardatan/graphql-mesh/blob/master/examples/json-schema-example/tests/json-schema-example.test.js). 

- Set up jest and related config
- Tests added:
  - `tests/UnifiedSchema.test.ts`
  - `tests/ZipLinkQuery.test.ts` 

## Notes
I originally mocked data with the `mock` plugin, but it was pointed out that this doesn't test our resolvers.  I've switched approaches to mock data retrieval methods called by the resolvers.  But I created a branch with the mock store approach, in case we want to use that in the future: https://github.com/chanzuckerberg/czid-graphql-federation-server/tree/testing-mock-store-example

## Tests
This PR adds automated tests

[CZID-9015]: https://czi-tech.atlassian.net/browse/CZID-9015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


json-schemas/amrWorkflowResults.json
```diff
@@ -33,7 +33,7 @@
                 }
             }
         },
-        "report_table_data": {
+        "amr_hit": {
```
jfoo1984, line 36: Align to schema returned by resolver.  This was preventing the tests from running, because the schema could not be generated

json-schemas/mngsWorkflowResults.json
```diff
@@ -104,7 +104,7 @@
                         "type": "object",
                         "properties": {
                             "tax_id": {
-                                "type": "integer"
+                                "type": "string"
```
jfoo1984, line 107: Align to schema returned by resolver.  This was preventing the tests from running, because the schema could not be generated

.github/workflows/conventional-commits.yml
```diff
@@ -12,4 +12,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: chanzuckerberg/github-actions/.github/actions/conventional-commits@v1.3.2
+      - uses: chanzuckerberg/github-actions/.github/actions/conventional-commits@main
```
jfoo1984, line 15: Since #61 was closed in favor of this PR, I am pulling this change in that was there.

.github/workflows/jest-unit-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libcurl4-openssl-dev
+      - name: Install package dependencies
+        run: |
+          npm ci --verbose --no-optional
+      - name: Run tests
+        run: |
+          npm test
```
jfoo1984, line 23: I tried to get these tests running with docker, but the tests are failing, and I couldn't get a clear indication as to what was causing the failure.  So I'm just going to leave it as running on the system.  

Since these are unit tests where a lot of data is mocked, I think this is acceptable, but if we want to pursue this in the future, here are the commands that have to be added before running tests

```
      - name: Create Docker network
        run: |
          docker swarm init
          npmdocker network create czidnet --attachable --driver overlay
      - name: Build Docker container for graphql mesh
        run: |
          make build
```

.github/workflows/deploy-staging.yml
```diff
@@ -1,7 +1,6 @@
 on:
-  push:
-    branches:
-      - main
+  release:
+    types: [published]
```
jfoo1984, line 3: this is in preparation for changing the release cadence.  stop pushing every PR to staging

rainandbare:
I like it Jerry! Great work! 

I couldn't quite make the Ziplink test fail on its own unless I deleted it from the yaml so maybe we would have to write more specific tests?  But even having the unified schema test is a logical and important safe guard. 

Anything we can do about these messages? 
Screenshot 2024-01-25 at 10 51 09 AM

jfoo1984:
I was able to remove the mock plugin, and instead, use jest mocking of our `httpUtil` function to actually test the resolvers instead of bypassing them.  This is probably a better test pattern we should use, given how much code we have in resolvers

The warnings are a type issue in the mock plugin, which I've raised with The Guild team.  With the removal of the mock plugin though, the warnings should no longer appear

jfoo1984:
I verified locally that changing code in the `ZipLink` resolver (to return a different response) does break the test.

rainandbare (approved):
I think this looks great Jerry! Are you worried that it is too much work per test?

jfoo1984:
> I think this looks great Jerry! Are you worried that it is too much work per test?

Thanks for reviewing @rainandbare!  Are you referring to added engineering work to write these tests?  My thought is that the amount of work to add these tests is inline with what I would expect for writing unit tests.  It does add overhead per feature we develop, but I think it's worth it for future maintainability of the system.  Hopefully, additional tests can follow the same patterns, and require less lift than this one did.

PR #71, opened 2024-01-24
feat: Federate the samplesAggregate NextGen query field

bchu1:
# Pull Request

## JIRA Ticket

[CZID-8821]

## Description

I've separated the fields that currently exist in NextGen (`where`) from those that are needed by the current Rails endpoint that don't have a NextGen equivalent and will need to exist until we get off of dual write (`notNextGen`).

## Tests

Endpoint is not used yet.

[CZID-8821]: https://czi-tech.atlassian.net/browse/CZID-8821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


utils/httpUtils.ts
```diff
@@ -1,78 +1,93 @@
...
-export const formatUrlParams = (params: any) => {
-    const paramList = Object.entries(params)
-        .filter(([_, value]) => value !== undefined)
-        .map(([key, value]) => `${key}=${value}`);
-    if (paramList.length === 0) {
-        return "";
-    }
-    return "?&" + paramList.join("&");
-}
+/**
```
jfoo1984, line 3: I know this PR wasn't the original place where these resolvers were created, but any chance we could add some unit tests for the resolvers that it touches (similar to [the ZipLink test](https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/112930adf4dd3921427a794b33dfdc9536bc57a9/tests/zipLinkQuery.test.ts#L12))?  If there is a sample response for these resolvers, we might be able to just parse that and stub the http call response with it.

resolvers.ts
```diff
@@ -77,42 +77,43 @@ export const resolvers: Resolvers = {
     },
     SampleMetadata: async (root, args, context, info) => {
       const url = `/samples/${args.sampleId}/metadata`;
-      const urlWithParams = args?.input?.pipelineVersion ? url + `?pipeline_version=${args?.input?.pipelineVersion}` : url;
-      const res = await get(
-        urlWithParams,
-        args,
-        context
-      );
+      const urlWithParams = args?.input?.pipelineVersion
```
jfoo1984, line 80: I know this PR wasn't the original place where these resolvers were created, but any chance we could add some unit tests for the resolvers that it touches (similar to [the ZipLink test](https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/112930adf4dd3921427a794b33dfdc9536bc57a9/tests/zipLinkQuery.test.ts#L12))?  If there is a sample response for these resolvers, we might be able to just parse that and stub the http call response with it.

PR #85, opened 2024-01-30
feat: Federate the samples NextGen field

bchu1:
# Pull Request

## JIRA Ticket

[CZID-9199]

## Description

This copies over the NextGen API while making some changes that haven't made it into NextGen's schema yet (e.g. `notes` is currently called `description`) and also including fields that are not part of NextGen but are required by the current Rails endpoint (see the `todoRemove` field).

Also enables `formatUrlParams()` to handle array query params of format `param[]=123&param[]=456`.

## Tests

Endpoint is not used.

[CZID-9199]: https://czi-tech.atlassian.net/browse/CZID-9199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.meshrc.yaml
```diff
@@ -104,17 +104,11 @@ sources:
             headers:
               Cookie: "{context.headers['cookie']}"
           - type: Query
-            field: Samples
-            path: /pub/{args.snapshotLinkId}/samples/{args.sampleId}.json
```
ehoops-cz, line 108: Just checking - does this still work if we need to include the snapshotLinkId?

bchu1: I don't think this GraphQL endpoint was even being used. `snapshotLinkId` is only used for short read MNGS, so I didn't port that over to the new endpoint.

resolvers.ts
```diff
@@ -193,27 +194,143 @@ export const resolvers: Resolvers = {
...
+                  nucleicAcid: sampleMetadata?.nucleotide_type,
+                  protocol: inputs?.wetlab_protocol,
+                  medakaModel: inputs?.medaka_model,
+                  technology: inputs?.technology,
+                  consensusGenomes: {
+                    edges: [
+                      {
+                        node: {
+                          taxon: {
+                            name: inputs?.taxon_name,
```
ninabernick, line 303: This doesn't exactly match the processing that happens in `processConsensusGenomeWorkflowRun` and `processRawWorkflowRun` -- are there fields that are unused in those old processing functions?

resolvers.ts
```diff
@@ -193,27 +194,143 @@ export const resolvers: Resolvers = {
...
+          notes: sampleInfo?.sample_notes,
+          collectionLocation: sampleMetadata?.collection_location_v2,
+          sampleType: sampleMetadata?.sample_type,
+          waterControl: sampleMetadata?.water_control,
+          hostTaxon: {
+            name: sampleInfo?.host_genome_name,
+          },
+          collection: {
+            name: sample?.project_name,
+            public: Boolean(sampleInfo?.public),
```
ninabernick, line 266: do we not need the collection id?

bchu1: For displaying the sample rows, it seems like we don't

resolvers.ts
bchu1: There are a few, like from a quick look `userIdWhoInitiatedWorkflowRun` is one. Did you see anything I missed? They should mostly be covered.

ninabernick: `taxonId` is the only one I think -- creation source looks like it was moved over to the workflow query?

bchu1: Yeah, I didn't bother querying that one cuz it's not used by the TableRenderers.ts

.meshrc.yaml
```diff
@@ -104,17 +104,11 @@ sources:
...
-            method: GET
-            responseSchema: ./json-schemas/samples.json
-            responseTypeName: Samples
-          - type: Query
-            field: Samples
-            path: /pub/{args.snapshotLinkId}/samples/index_v2.json?projectId={args.projectId}
+            field: samples
             method: GET
-            responseSchema: ./json-schemas/samples.json
-            responseTypeName: Samples
```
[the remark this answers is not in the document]
bchu1: 👍

resolvers.ts
```diff
@@ -193,27 +194,143 @@ export const resolvers: Resolvers = {
...
-        }
-        return formatSample(sample);
-      } else if (args.projectId) {
-        const { samples } = await get(
-          `/samples/index_v2.json?projectId=${args.projectId}&snapshotShareId=&basic=true`,
-          args,
-          context
-        );
-        return formatSamples(samples);
+    samples: async (root, args, context) => {
```
rainandbare, line 197: Any reason you went with lower case rather than upper case? Closer to next gen functionality?

resolvers.ts
```diff
@@ -77,37 +77,38 @@ export const resolvers: Resolvers = {
     },
     SampleMetadata: async (root, args, context, info) => {
       const url = `/samples/${args.sampleId}/metadata`;
-      const urlWithParams = args?.input?.pipelineVersion ? url + `?pipeline_version=${args?.input?.pipelineVersion}` : url;
-      const res = await get(
-        urlWithParams,
-        args,
-        context
-      );
+      const urlWithParams = args?.input?.pipelineVersion
```
rainandbare, line 80: Is this all just formatting in MetadataFields?

rainandbare (approved):

resolvers.ts
bchu1: Exactly, otherwise when we start exposing NextGen directly we'll have to change the cases

resolvers.ts
bchu1: Should all just be formatting from Prettier

PR #86, opened 2024-01-31
feat: add BulkDownloadCSV query

rainandbare:
# Pull Request

## JIRA Ticket

[CZID-9313](https://czi-tech.atlassian.net/browse/CZID-9313)

## Description

Part 1 of 2 of Federating /bulk_downloads. This query will be used for the "sample_metadata" and "consensus_genome_overview" download types. 

## Notes

I am trying to structure this so that I can use it on the front end before the backend has made any changes. As of now, it returns { url: "success" } if it has successfully kicked off the pipeline and I can use the success state to trigger the notification on the front end.  

After the CSV rails and NextGen integrations are in place I will be able to us the return value of {url: [URLVALUE]} to directly download a file on the front end.

## Tests

**TODO:** I would like some input on how to test for a failed BulkDownload creation currently on Rails. 

```
query BulkDownloadCSVQuery {
  BulkDownloadCSV(
    input: {
      authenticityToken: [AUTH_TOKEN_GOES_HERE],
      downloadType: "consensus_genome_overview", 
      includeMetadata: true,
      workflow: "consensus_genome", 
      workflowRunIds: [1991, 2007]
    }
  ) {
    error
    url
  }
}
```

[CZID-9313]: https://czi-tech.atlassian.net/browse/CZID-9313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.meshrc.yaml
```diff
@@ -39,6 +39,13 @@ sources:
             responseTypeName: BulkDownload
             headers:
               Cookie: "{context.headers['cookie']}"
+          - type: Query
+            field: BulkDownloadCSV
```
jfoo1984, line 43: Would it be more desirable to have separate graphQL calls for different download types (CG overview and metadata)?  They can call common functions in the front / back end, but I feel like it's more explicit.  The current bulk download endpoint has a lot of branching logic, and I feel like it'd be nice to move away from that...

resolvers.ts
```diff
@@ -40,6 +40,38 @@ export const resolvers: Resolvers = {
...
+      );
+      if (res.status !== 200){
+        return {
+          url: null,
+          error: res.statusText,
+        }
+      }
+      return {
+        url: res.csvUrl ? res.csvUrl : "success",
+      }
```
jfoo1984, line 73: I was planning to send the file back directly, using [the Rails function 'send_data`](https://api.rubyonrails.org/classes/ActionController/DataStreaming.html#method-i-send_data).  We do this in a few other places in Rails already.  A couple of questions around this though

1.  How is the entities service going to return the file?  A URL only works if we are writing the file somewhere, but my impression is that we weren't.
2. I'm not sure how we represent and federate a file response.  Depending on how that works, do we even need a resolver function for this if we don't need to change the response?  Though I guess we will need it eventually to fetch CG metrics from nextGen and pass them to Rails.

.meshrc.yaml
rainandbare: Without Sample Metadata we don't have to worry about this now right? I can change the name to something more focused - `BulkDownloadOverview`?

resolvers.ts
rainandbare: For future reference - we decided that we would send back a json string and the front end would be responsible for turning it into a CSV file and downloading.

PR #88, opened 2024-02-01
feat: [CZID-9279] Add entities to graphql sources

rzlim08:
# Pull Request

## JIRA Ticket

[CZID-9279](https://czi-tech.atlassian.net/browse/CZID-9279)

## Description

* Add entities as a data source 
* Does a passthrough of the authorization token
* Tested locally and on dev deploy

[CZID-9279]: https://czi-tech.atlassian.net/browse/CZID-9279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.happy/terraform/envs/dev/main.tf
+++ b/.happy/terraform/envs/dev/main.tf
@@ -6,8 +6,11 @@ module "stack" {
   deployment_stage = "dev"
   stack_prefix     = "/${var.stack_name}"
   k8s_namespace    = "czid-dev-happy-happy-env"
+  app_name         = var.app
   additional_env_vars = {
     API_URL = "https://sandbox.czid.org"
+    NEXTGEN_ENTITIES_URL = "http://ryan-test-entities.czid-dev-happy-happy-env.svc.cluster.local:8008"
+    NEXTGEN_WORKFLOWS_URL = "http://workflows.czidnet"
   }
   services = {
     gql = {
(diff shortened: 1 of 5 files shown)


.happy/terraform/envs/dev/main.tf
```diff
@@ -6,8 +6,11 @@ module "stack" {
   deployment_stage = "dev"
   stack_prefix     = "/${var.stack_name}"
   k8s_namespace    = "czid-dev-happy-happy-env"
+  app_name         = var.app
   additional_env_vars = {
     API_URL = "https://sandbox.czid.org"
+    NEXTGEN_ENTITIES_URL = "http://ryan-test-entities.czid-dev-happy-happy-env.svc.cluster.local:8008"
```
jgadling, line 12: Have you poked the Happy team about how to update these values on the fly? I think `happy config` is one possible option, but it would be nice to be able to launch different remote-dev stacks with different configurations.

jgadling (approved):
I think we need to work out how to make these values configurable soon, but this is a start!

.happy/terraform/envs/dev/main.tf
jfoo1984: I guess I never changed this for dev, but for other envs (sandbox, staging prod), I've been putting ENV vars into Github.  I think it's probably fine for dev to leave it here though, since we generally won't be deploying that  via Github.

PR #89, opened 2024-02-02
feat: Federate workflowRuns endpoint

bchu1:
# Pull Request

## JIRA Ticket

[CZID-9271]

## Description

Next: Query both Workflows and Entities to get the complete list of all workflow run IDs, which the FE relies on to grabbing which metadata fields to display and triggering an NextClade or bulk download action on all runs even if they haven't been lazy loaded.

Also includes some improvements to the `samples` test.

## Tests

Not used.

[CZID-9271]: https://czi-tech.atlassian.net/browse/CZID-9271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.meshrc.yaml
```diff
@@ -144,6 +144,12 @@ sources:
             requestSample: {"selectedIds":[28114,28151],"workflow":"short-read-mngs","authenticityToken":"token"}
             responseSchema: ./json-schemas/validateUsersCanDeleteObjects.json
             responseTypeName: ValidateUserCanDeleteObjects
+          - type: Query
+            field: workflowRuns
+            path: /workflow_runs.json
+            method: POST
+            requestSchema: ./json-schemas/workflowRunsRequest.json
+            responseSchema: ./json-schemas/workflowRunsResponse.json
```
jfoo1984, line 152: I think @saq7 has been working on federating this endpoint in #92 as well.  we should probably see if we can get all the changes in one PR to avoid conflicts?

bchu1: From what we discussed offline, should we just merge this first? We can decide if we want to build off this and conditionally hit the other Rails endpoint, or just use a different federation endpoint for now?

jfoo1984 (approved):

PR #91, opened 2024-02-02
feat: Federate CreateBulkDownload Endpoint

rainandbare:
# Pull Request

## JIRA Ticket

[CZID-XXXX]

## Description

## Notes

## Tests


package.json
```diff
@@ -6,7 +6,7 @@
     "start": "mesh start",
     "build": "mesh build",
     "dev": "mesh dev",
-    "test": "mesh build && MOCKING_ENABLED=1 jest"
+    "test": "mesh build && jest"
```
jfoo1984, line 9: 👍

.meshrc.yaml
```diff
@@ -39,13 +39,20 @@ sources:
             responseSchema: ./json-schemas/snapshotBackground.json
             responseTypeName: Background
           - type: Query
-            field: BulkDownload
+            field: BulkDownloadDetails
```
jfoo1984, line 42: was this conflicting with some other field?  I think this will actually be the entity name in nextGen (`BulkDownload`) so would we want to leave it as is?

jfoo1984: One issue I can see would be that for graphQL, the `BulkDownload` schema field would accept an array of IDs, whereas our Rails endpoint is setup to just retrieve one, so we'd have to handle that in the resolver or update the Rails endpoint

resolvers.ts
```diff
@@ -40,6 +40,31 @@ export const resolvers: Resolvers = {
...
+        params: {
+          include_metadata: { value: args?.input?.includeMetadata }, 
+        sample_ids: {
+          value: args?.input?.workflowRunIds
+        }, 
+        workflow: {
+          value: "consensus_genome"
+        }
+        },
+        workflow_run_ids: args?.input?.workflowRunIds,
```
jfoo1984, line 56: nit: could we destructure `args?.input` to make this a bit more succinct?  Also, it seems like if `args.input` (or any of the fields) do not exist, would we want to make the API call even?

tests/ZipLinkQuery.test.ts
```diff
@@ -20,7 +20,7 @@ describe("ZipLink Query", () => {
     query = getZipLinkExampleQuery();
   });
 
-  describe.only("ZipLink with url", () => {
+  describe("ZipLink with url", () => {
```
jfoo1984, line 23: thanks for re-enabling and fixing these tests!

utils/httpUtils.ts
```diff
@@ -64,6 +65,9 @@ export const postWithCSRF = async (url: string, body: any, args: any, context: a
             body: JSON.stringify(body),
           });
         checkForLogin(response?.url);
+        if (fullResponse) {
```
jfoo1984, line 68: the default value of the parameter is a string.  I think that will technically evaluate to true, but should we make it a boolean instead?

.meshrc.yaml
rainandbare: This was not returning a singular Bulk Download. But a Bulk Download plus other details. I changed it to make room for a future singular Bulk Download.

tests/utils/ExampleQueryFiles.ts
```diff
@@ -10,3 +10,36 @@ const getZipLinkExampleQueryPath = () : string => {
 export const getZipLinkExampleQuery = () : string => {
   return readFileSync(getZipLinkExampleQueryPath(), { encoding: "utf8" });
 }
+
```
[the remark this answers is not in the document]
rainandbare: Sure - that would be great.

utils/httpUtils.ts
rainandbare: I had intended that the type of the parameter is a specific string "fullResponse". If you pass anything other than that specific string you will get a type error. It is easier to understand what is happening when you call the function when you use a full string. But you are right I should check for the exact string here. 

In the end, however I did not use it. I can take it away.

.meshrc.yaml
jfoo1984: ah I see now.  So this might just need to be a special snowflake endpoint for now, since it doesn't really fit with typical graphQL / nextGen schema?

rainandbare: For now! It gets used in BulkDownloadList so I will take a look again in the next mini-project!

json-schemas/bulkDownload.json
```diff
@@ -0,0 +1,99 @@
...
+          "type": "string"
+      },
+      "download_type": {
+          "type": "string"
+      },
+      "status": {
+          "type": "string"
+      },
+      "error_message": {
+          "type": "NoneType"
```
bchu1, line 17: whats the purpose of these `NoneType`s?

bchu1 (approved):

json-schemas/bulkDownload.json
rainandbare: Those are most often errors from the compiler thanks for pointing it out!

tests/utils/ExampleQueryFiles.ts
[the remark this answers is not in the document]
rainandbare: Could we take care of this in a different PR Jerry? So as not to block this one?

jfoo1984 (approved):

PR #96, opened 2024-02-08
feat: Add entities to sources

rzlim08:
* Lets try this again since I've added a PR in web app that defines the `node` id as `_id`


sources/entities-schema.graphql
```diff
@@ -0,0 +1,1961 @@
...
+  createPhylogeneticTree(input: PhylogeneticTreeCreateInput!): PhylogeneticTree!
+  updatePhylogeneticTree(input: PhylogeneticTreeUpdateInput!, where: PhylogeneticTreeWhereClauseMutations!): [PhylogeneticTree!]!
+  deletePhylogeneticTree(where: PhylogeneticTreeWhereClauseMutations!): [PhylogeneticTree!]!
+}
+
+"""An object with a Globally Unique ID"""
+interface Node {
+  """The Globally Unique ID of this object"""
+  _id: GlobalID!
+}
```
rainandbare, line 815: So Relay still refuses to compile with this type. But if we can get it to somehow be:
```
interface Node {
  """The Globally Unique ID of this object"""
  _id: GlobalID!
  id: ID!
}
```
It was able to compile.

rainandbare: Oh! Sorry - I didn't understand. With the PR from the web app all is grand. Thanks for that fix. 
✅

rainandbare (approved):
Thanks Ryan!

jfoo1984:
So the same `.meshrc.yml` is used in all environments.  Since there are no configurations for the staging and prod env vars, will we run into issues when deploying to those environments?  (I'm not sure if you tested what happens in sandbox if the `.meshrc` includes the entities service but does not set the entities URL env var).

Also, for this service, we've been configuring URLs using github environment variables / secrets, rather than setting them terraform.  could we switch to that pattern?  Unfortunately, that also means we need to update all of the deploy scripts to set the variables from Github (it might be time to create a reusable action for those things).

PR #99, opened 2024-02-12
chore: Add util function for retrieving enriched tokens

jfoo1984:
# Pull Request

## JIRA Ticket

[CZID-9251]

## Description

## Notes

## Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ services:
     restart: always
     environment:
       API_URL: http://web:3001
+      LEGACY_API_URL: http://web:3001
       ALLOWED_CORS_ORIGINS: "http://web:3001,http://localhost:3000"
       NEXTGEN_ENTITIES_URL: "http://entities.czidnet:8008"
       NEXTGEN_WORKFLOWS_URL: "http://workflows.czidnet:8042"
--- a/utils/enrichToken.test.ts
+++ b/utils/enrichToken.test.ts
@@ -1,0 +1,46 @@
+import { getEnrichedToken } from './enrichToken';
+
+describe('getEnrichedToken', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  const getContext = (cookieStr) => {
+    const mockedHeaders: unknown = {
+      get: jest.fn().mockReturnValue(cookieStr),
+    };
+
+    return {
+      request: {
+        headers: mockedHeaders as Headers,
+      },
+    };
+  };
+
+  it('should throw an error if czid_services_token is not found in cookie', async () => {
+    const context = getContext('abc=123');
+    await expect(getEnrichedToken(context)).rejects.toThrow('No czid_services_token found in cookie');
+  });
+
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -10,6 +10,7 @@ services:
     restart: always
     environment:
       API_URL: http://web:3001
+      LEGACY_API_URL: http://web:3001
```
jfoo1984, line 13: This is to make it more clear what this API is, as we are integrating APIs for new services.

utils/enrichToken.test.ts
```diff
@@ -0,0 +1,46 @@
+import { getEnrichedToken } from './enrichToken';
```
jfoo1984, line 1: Calling out that for this file, I put it in alongside the util file (in the same directory).  For our query tests, those are located in a specific tests folder, but I think it's generally easier to find the test if you locate it with the file it's testing.  It's a bit weird to have two different patterns, but any thoughts on this?

We could actually adopt this pattern by splitting our resolver code into separate ts files in a specific directory, and putting the query tests along them (but that's a nice to have for another day).

utils/enrichToken.ts
```diff
@@ -0,0 +1,47 @@
...
+    .split(';')
+    .map((keyValStr: string) => keyValStr.split('='))
+    .map((keyValArr: [string, string]) => ([
+      decodeURIComponent(keyValArr[0].trim()),
+      decodeURIComponent(keyValArr[1].trim()),
+    ]))
+    .find((keyValArr) => keyValArr[0] === "czid_services_token");
+
+  return tokenKeyVal ? tokenKeyVal[1] : null;
+}
```
valenzuelaomar, line 12: What should a cookieStr look like? This could be helpful context to have in a comment when reading this function so the reader can follow along with the .split() & .maps()

jfoo1984: These are standard HTTP cookies, so I don't think there's a need to add any specific documentation here.  I'll rename the variable to `httpCookieStr`, to be explicit about that.  The unit tests are also a good place to self-document things like this, so I'll update the unit tests to have mock cookies with multiple keys (which is closer to what they really look like).

docker-compose.yml
jfoo1984: Noted in the PR description, I decided not to switch the `API_URL` references for now, because that will require all deploy scripts to be updated.  I will do that in a separate PR, to avoid adding complexity here.

utils/callGraphqlSource.ts
```diff
@@ -0,0 +1,15 @@
...
+
+  // TODO: Implement the actual call to the graphql source, something like the below
+  // const resp = await fetch(url, {
+  //   method: "POST",
+  //   headers: {
+  //     Cookie: context.request.headers.get("cookie"),
+  //     "Content-Type": "application/json",
+  //     Authorization: `Bearer ${enrichedToken}`,
+  //   },
+  // );
```
jfoo1984, line 14: This comment is intentional.  @saq7 will replace this with the appropriate code when integrating this with the new core services

PR #102, opened 2024-02-12
feat: Federate sequencingReads and consensusGenomes

bchu1:
# Pull Request

## JIRA Ticket

[CZID-9351]
[CZID-9353]

## Description

These 2 endpoints will be used for retrieving the actual (paginated) sample rows. The will used the list of perfectly filtered IDs as input in NextGen.

`sequencingReads` can fully substitute for querying `samples` since it's a 1:1 relationship (can handle sorting on nested `Sample`s). It will also be the query that handles sorting by metadata fields, which (while Rails contains all metadata) will require the Federation layer to query a different Rails endpoint for the `Sample`s sorted by metadata.

`consensusGenomes` is necessary for sorting on `ConsensusGenome` fields. The query itself is not sufficient due to `Sample`s that don't have `ConsensusGenome`s, so the FE will have to stitch the result together with those `Sample`s.

Also updated the `samples` endpoint:

- Added inputs for the 2 separate queries we'll have to "`OUTER JOIN`" to filter by taxon.
- Fixed some of the `id` logic.

Separates out the test queries into their own files so that Omar can use them for his API diff testing. Also cleans up the example files util (we don't need a new function every test).

## Tests

Unused endpoints.

[CZID-9351]: https://czi-tech.atlassian.net/browse/CZID-9351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CZID-9353]: https://czi-tech.atlassian.net/browse/CZID-9353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tests/utils/ExampleQueryFiles.ts
+++ b/tests/utils/ExampleQueryFiles.ts
@@ -1,45 +1,22 @@
-import { readFileSync } from 'fs';
+import { readFileSync } from "fs";
 import { join } from "path";
 
-const exampleQueriesDir = "../../example-queries";
-
-const getZipLinkExampleQueryPath = () : string => {
-  return join(__dirname, `${exampleQueriesDir}/zip-link-query.graphql`);
-};
-
-export const getZipLinkExampleQuery = () : string => {
-  return readFileSync(getZipLinkExampleQueryPath(), { encoding: "utf8" });
+export function getExampleQuery(graphqlFile: string): string {
+  return readFileSync(
+    join(__dirname, `../../example-queries/${graphqlFile}.graphql`),
+    {
+      encoding: "utf8",
+    }
+  );
 }
 
-
-const getBulkDownloadCGOverviewExampleQueryPath = () : string => {
-  return join(__dirname, `${exampleQueriesDir}/bulk-download-cg-overview-query.graphql`);
-};
-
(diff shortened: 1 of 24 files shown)


valenzuelaomar (approved):
Thanks for extracting the queries out into example-queries!

tests/utils/ExampleQueryFiles.ts
```diff
@@ -1,45 +1,30 @@
...
-export const getBulkDownloadCGOverviewResponse = () : string => {
-  return readFileSync(getBulkDownloadCGOverviewResponsePath(), { encoding: "utf8" });
-} 
-
-const getCreateBulkDownloadExampleMutationPath = () : string => {
-  return join(__dirname, `${exampleQueriesDir}/create-bulk-download-query.graphql`);
+/**
+ * @param graphqlFile file name (excludes .graphql file extension)
+ * @returns example query as a string
+ */
```
jfoo1984, line 7: thanks for DRY-ing the code here!

PR #108, opened 2024-02-14
feat: federate WorkflowsAggregate endpoint

ninabernick:
# Pull Request

## JIRA Ticket

[CZID-9306]

## Description

Federate an endpoint to aggregate workflow counts by project and workflow type to use in the projects view to display X mNGS | Y CG | Z AMR. Note that filters apply to these counts for whatever reason, so in order to maintain those counts having filters applied in NextGen, we want to pass workflow run ids into the request to group by project and workflow.

Rails:
- hits `projects.json` a second time with all the filters applied. Potential to refactor this endpoint to remove some extraneous queries and only get the aggregates, but this endpoint is kind of a mess and I'd like to avoid touching it.
- uses all the filters specified in `todoRemove` to filter down projects and samples

NextGen:
- will hit a new aggregate endpoint with workflow run ids for each type and receive an aggregate by project id and workflow type

## Notes

Very low confidence in the shape of the response from nextgen

## Tests

Hit endpoint from `localhost:3000/graphqlfed` and verified projects correctly aggregated and returned. + unit test for resolver.

[CZID-9306]: https://czi-tech.atlassian.net/browse/CZID-9306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.meshrc.yaml
```diff
@@ -150,6 +150,13 @@ sources:
             method: POST
             requestSchema: ./json-schemas/workflowRunsRequest.json
             responseSchema: ./json-schemas/workflowRunsResponse.json
+          - type: Query
+            field: WorkflowRunsAggregate
```
bchu1, line 154: should we lowercase this since that's what it looks like in NextGen?

json-schemas/workflowRunsAggregateRequest.json
```diff
@@ -0,0 +1,130 @@
...
+                "taxon": {
+                    "type": "array",
+                    "items": {
+                        "type": "integer"
+                    }
+                },
+                "search": {
+                    "type": "string"
+                },
+                "orderBy": {
```
bchu1, line 121: we can get rid of the order bys

json-schemas/workflowRunsAggregateRequest.json
```diff
@@ -0,0 +1,130 @@
+{
+    "type": "object",
+    "properties": {
+        "workflowRunIds": {
+            "type": "object",
+            "properties": {
+                "short-read-mngs": {
+                    "type": "array",
+                    "items": {
+                        "type": "integer"
```
bchu1, line 10: might be good to just start sending `string`s now, since NextGen will expect them

json-schemas/workflowRunsAggregateRequest.json
```diff
@@ -0,0 +1,130 @@
+{
+    "type": "object",
+    "properties": {
+        "workflowRunIds": {
```
bchu1, line 4: I was guessing it would something like `where { id { _in } }` but totally ok with keeping it like this for now since the API hasn't actually been released 😄

resolvers.ts
```diff
@@ -459,6 +460,45 @@ export const resolvers: Resolvers = {
...
       );
     },
+    WorkflowRunsAggregate: async (root, args, context, info) => {
+      const input = args.input;
+      
+      const { projects } = await get("/projects.json" + 
+        formatUrlParams({
+          projectId: input?.todoRemove?.projectId,
+          domain: input?.todoRemove?.domain,
+          limit: TEN_MILLION, // do we need pagination? or single request?
```
bchu1, line 470: maybe we can paginate if we see some performance issues? Otherwise I'm fine trying out all in one go for now!

json-schemas/workflowRunsAggregateRequest.json
```diff
@@ -0,0 +1,111 @@
+{
+    "type": "object",
+    "properties": {
+        "where": {
+            "type": "object",
+            "properties": {
+                "id": {
```
ninabernick, line 7: I only have one field for workflow run ids. We will probably need to change this once we start querying more than one workflow for nextgen but that's a problem for later.

bchu1: 1 field should still work right? Since each workflow run can only belong to 1 workflow anyways? Unless I'm wrong

bchu1 (approved):

json-schemas/workflowRunsAggregateRequest.json
ninabernick: I think it just depends how we implement it from the frontend, e.g. make 4 agg queries (one for each workflow) or join all the workflow run ids together.

PR #110, opened 2024-02-16
feat: Feature Flag based routing

jfoo1984:
# Pull Request

## JIRA Ticket

[CZID-8731]

## Description
Reopening #44

## Notes

## Tests


jfoo1984:
Re-adding comments from #44 

@saq7 Can we update the unit tests before merging this? 

Also, could we make sure to fill out the PR description to make it easier for someone looking back in the future to understand what this change was intended to do?

utils/httpUtils.ts
```diff
@@ -91,3 +147,42 @@ export const notFound = (message: string) => {
...
+  } catch (e) {
+    return Promise.reject(e.response);
+  }
+};
+
+export const getFeatureFlagsFromRequest = (context) => {
+  return context.request.headers.get("readFromNextGen");
+};
+
+export const isNextGenEnabled = async (context) => {
```
jfoo1984, line 170: nit: could we rename this to something like shouldReadFromNextGen? The current name may be too generic, in case we end up needing to creating / check any other next gen related feature flags.

Specifically, I could imagine that we end up needing additional feature flags around writes and/or reading for particular features.

utils/httpUtils.ts
```diff
@@ -40,15 +60,35 @@ export const get = async (url: string, args: any, context: any) => {
...
+        method: "POST",
+        headers: {
+          Cookie: context.request.headers.get("cookie"),
+          "Content-Type": "application/json",
+          "X-CSRF-Token": args?.input?.authenticityToken,
+          Authorization: `Bearer ${czidServicesToken}`,
+        },
+        body: JSON.stringify(query),
+      });
+      return response;
```
jfoo1984, line 78: (reposting [previous comment](https://github.com/chanzuckerberg/czid-graphql-federation-server/pull/44#discussion_r1491386917))

would it be viable to put this code in the `callGraphqlSource` util function instead, and call that util function here?  This util file is getting fairly long and complex, so it'd be nice to extract the chunks that make sense.

https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/010967bd3011dd1544a6bad984f4e64e2f3f0b08/utils/callGraphqlSource.ts#L3

utils/httpUtils.ts
```diff
@@ -67,19 +107,35 @@ export const postWithCSRF = async (
...
+        headers: {
+          Cookie: context.request.headers.get("cookie"),
+          "Content-Type": "application/json",
+          "X-CSRF-Token": args?.input?.authenticityToken,
+          Authorization: `Bearer ${czidServicesToken}`,
+        },
+        body: JSON.stringify(query),
+      });
+      return await response.json();
+      // next gen details
```
jfoo1984, line 125: same comment as above: https://github.com/chanzuckerberg/czid-graphql-federation-server/pull/110/files#r1492821002

utils/httpUtils.ts
jfoo1984: I had another thought here, that might be worth adding a comment about.  All the calls that Relay makes to graphQL mesh are done via POST, so it feels a bit weird that within the get method call, we are calling this.  

I understand why that it's necesary, because otherwise we would need to branch in every resolver based on the feature flag, which is not ideal.  But I'm wondering if that's worth a one line comment to explain in case someone is confused in the future.

utils/httpUtils.ts
```diff
@@ -22,16 +23,35 @@ export const formatUrlParams = (params: { [s: string]: unknown }) => {
...
+        method: "POST",
+        headers: {
+          Cookie: context.request.headers.get("cookie"),
+          "Content-Type": "application/json",
+          "X-CSRF-Token": args?.input?.authenticityToken,
+          Authorization: `Bearer ${czidServicesToken}`,
+        },
+        body: JSON.stringify(query),
+      });
+      return await response.json();
```
jfoo1984, line 42: same comment below, extract to shared method in util.

utils/httpUtils.ts
```diff
@@ -40,15 +60,35 @@ export const get = async (url: string, args: any, context: any) => {
...
-      headers: {
-        Cookie: context.request.headers.get("cookie"),
-        "Content-Type": "application/json",
-      },
-    });
-    return response;
+    const nextGenEnabled = await isNextGenEnabled(context);
+    if (nextGenEnabled) {
+      // next gen details
+      const czidServicesToken = await getEnrichedToken(context);
```
jfoo1984, line 66: I wanted to clarify this variable name.  The czid services token is different from the enriched token.  That token is a simple user id token, which we send to rails to "exchange" for the enriched token.  I think we should use the proper names here to avoid confusion on what token is being sent to nextgen.

That said, if we extract this code to the shared util function, we don't really need to worry about it here, and the stub method already uses the accurate name: https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/010967bd3011dd1544a6bad984f4e64e2f3f0b08/utils/callGraphqlSource.ts#L4

.meshrc.yaml
```diff
@@ -1,18 +1,25 @@
...
+  #       source: ./sources/nextgen-entities-schema.graphql
+  #       operationHeaders:
+  #         Cookie: "{context.headers['authorization']}"
+  # - name: NextGenWorkflows
+  #   handler:
+  #     graphql:
+  #       endpoint: ${NEXTGEN_WORKFLOW_API_URL}/graphql
+  #       source: ./sources/nextgen-workflows-schema.graphql
+  #       operationHeaders:
+  #         Cookie: "{context.headers['authorization']}"
```
jfoo1984, line 22: should these be uncommented?

jfoo1984:
Will branching based on feature flag in the httpUtil functions work in cases where we may need to call the workflows service?  Most of our existing endpoints probably go to entities, but I think anything that kicks off a workflow will need to be routed to that service right?

utils/httpUtils.ts
rainandbare: I don't know if this is going to work the way we want because in the custom resolvers we are already massaging the Rails return to fit the PREDICTED NextGen. In the short term, it might be easier to use the feature flag in each resolver to output the ACTUAL NextGen response and map it to what the front end now expects, rather than changing the front end again right now?

What do you think?

rainandbare:
> Will branching based on feature flag in the httpUtil functions work in cases where we may need to call the workflows service? Most of our existing endpoints probably go to entities, but I think anything that kicks off a workflow will need to be routed to that service right?

I have a case in bulkDownloads where I would like to be able to call rails too. Depending on the download type.

jfoo1984:
> > Will branching based on feature flag in the httpUtil functions work in cases where we may need to call the workflows service? Most of our existing endpoints probably go to entities, but I think anything that kicks off a workflow will need to be routed to that service right?
> 
> I have a case in bulkDownloads where I would like to be able to call rails too. Depending on the download type.

@saq7 and I spoke about this, and our initial thought was we could punt this and figure it out when we actually have a need to call the workflow service.  However, this bulk download case sounds more immediate, so I think we need to re-consider where we do this branching.  

It's repetitive, but having each resolver handle which service it calls may be the way to go for clarity.  Or if the util functions in this PR work most of the time, maybe we can just have some of the resolvers have their own logic.  We could probably still reuse code in the util functions as long as we split them out into shared functions.

@saq7 is this something you can work through next week?

PR #130, opened 2024-02-28
chore: Temporarily rename nextgen workflowRuns top level field to tempWorkflowRuns

jgadling:
I know ideally we'd like to have a *copy* type of operation so we can seamlessly move this back where it needs to be long-term, but so far the transform plugs I've been working with don't quite seem to do that 😱 ... this should get us unblocked for now and buy some time to figure out the better solution.

### DIFF
--- a/.meshrc.yaml
+++ b/.meshrc.yaml
@@ -290,16 +290,6 @@ sources:
         - namingConvention:
             typeNames: camelCase
             fieldNames: camelCase
-transforms:
-  - rename:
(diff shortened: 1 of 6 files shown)


bchu1:
LGTM, do you mind generating the snapshot so we can see what the schema looks like?

bchu1 (approved):

bchu1:
Thanks! schema looks 💚

PR #134, opened 2024-02-29
chore: Increase header size limit.

jgadling:
# Pull Request

## JIRA Ticket

[CZID-XXXX]

## Description

## Notes

## Tests


Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM node:18.16.0
 
 WORKDIR /usr/src/app
+ENV UWS_HTTP_MAX_HEADERS_SIZE=8192
```
jfoo1984, line 4: All the other env vars are being set via github variables / secrets, though since I don't think this value should vary between environments, it probably makes more sense to do this one in the Dockerfile.  The only drawback I can see is you need a code change to update this (which isn't a big deal either).

In our nginx configs, we've set the max header size to 24k, though that was probably just an arbitrary large value.  Should we make those settings consistent with this, if 8k is sufficient?

jfoo1984 (approved):
looks good to me, thanks for doing this.

Dockerfile
jgadling: I'lI update this to 24k to match nginx. Also, this value can be overridden by service configurations, this is just a default.

PR #137, opened 2024-03-01
feat: Create Connection between Fed Server and Next Gen 

rainandbare:
# Pull Request

## JIRA Ticket

[CZID-8731]

## Description

This PR pulls in the connection code to main so that we are all synced up between sandbox and main. 

This will allow us to work together on creating connections for individual resolvers.

There are a lot of formatting changes in this PR unfortunately but the main work is done in `httpUtils`. I changed the `get` function to receive an object and pulled out `getFromRails` and `fetchFromNextGen` into their own functions so that we can call them directly in the resolvers if needed.

We can either call `fetchFromNextGen` using the passed through `context.params.query` with a formatting function that pulls out the 'fed' prefix and the `input: {}` wrapper or we can pass an optional custom query if we need something specific in the resolvers.

## Tests

Updated unit tests to cover new fedConsensusGenome functionality
Smoke tested with main branch of `czid-web` to ensure no changes in functionality


tests/FormatNextGenQuery.test.ts
```diff
@@ -0,0 +1,35 @@
...
+describe("FormatFedQuery:", () => {
+  describe("formatFedQuery:", () => {
+    it("Removes fed prefix from the top level query", () => {
+      expect(formatFedQueryForNextGen("query fedConsensusGenome")).toBe("query consensusGenome");
+    });
+
+    // it("It does not remove fed from lower in the query", () => {
+    //   expect(
+    //     formatFedQueryForNextGen({
+    //       param1: 123,
```
rainandbare, line 12: I would like to complete these two pieces of functionality for this function but I thought it would be good to get to main first so that everyone can use this connection code.

utils/httpUtils.ts
```diff
@@ -72,3 +71,111 @@ export const notFound = (message: string) => {
...
+    splitQueryOnFed[1] = splitQueryOnFed[1].slice(1);
+    splitQueryOnFed.splice(1, 0, firstLetterLowerCase);
+    cleanQuery = splitQueryOnFed.join("");
+  }
+
+  // remove input object from variables
+  const finishedQuery = cleanQuery
+    .replace("input: {", "")
+    .replace("}}", "}")
+    // TODO: (suzette 02/27/24) FIX THESE HACKS TO MAKE THIS FUNCTION MORE UNIVERSAL
```
rainandbare, line 103: This function is a little lame at the moment. I am hoping it can mature as we add more connections.

tests/FormatNextGenQuery.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+
+    // it("It can handle a list of bespoke swaps", () => {
+    //   expect(
+    //     formatFedQueryForNextGen({
+    //       param1: [123, 456],
+    //       param2: [],
+    //     })
+    //   ).toBe("?&param1[]=123&param1[]=456");
+    // });
```
ehoops-cz, line 32: is this meant to be commented out?

tests/FormatNextGenQuery.test.ts
```diff
@@ -0,0 +1,40 @@
+import { formatFedQueryForNextGen } from "../utils/httpUtils";
+
+describe("FormatFedQuery:", () => {
+  describe("formatFedQuery:", () => {
+    it("Removes fed prefix from the top level query", () => {
+      expect(
+        formatFedQueryForNextGen(
```
ehoops-cz, line 7: Just to make sure I understand correctly - we are using fedQueryObject for things that we want to match nextGen, but slightly don't match.  This will make it easier to modify the query if `nextGenEnabled === true` ?

utils/httpUtils.ts
```diff
@@ -1,56 +1,55 @@
...
+      if (!serviceType) {
+        console.error("You must pass a service type to call next gen");
+        throw new Error("You must pass a service type to call next gen");
+      } else {
+        return fetchFromNextGen({ args, context, serviceType, fullResponse, customQuery });
+      }
+    } else {
+      if (!url) {
+        console.error("You must pass a url to call rails");
+        throw new Error("You must pass a service type to call next gen");
```
ehoops-cz, line 31: This one should match the line above?

utils/httpUtils.ts
```diff
@@ -72,3 +71,114 @@ export const notFound = (message: string) => {
...
   });
 };
+
+export const getFeatureFlagsFromRequest = context => {
+  return context.request.headers.get("x-should-read-from-nextgen");
+};
+
+export const shouldReadFromNextGen = async context => {
+  let shouldReadFromNextGen = getFeatureFlagsFromRequest(context);
+  if (shouldReadFromNextGen === true || shouldReadFromNextGen === "true" || shouldReadFromNextGen === "True") {
```
ehoops-cz, line 81: 😂

utils/httpUtils.ts
```diff
@@ -72,3 +71,114 @@ export const notFound = (message: string) => {
...
+    cleanQuery = splitQueryOnFed.join("").concat(splitOnFirstFed[1]);
+  }
+
+  // remove input object from variables
+  const finishedQuery = cleanQuery
+    .replace("input: {", "")
+    .replace("}}", "}")
+    // apply any specific type switches that need to be made - these can be passed in from the resolver
+    .replace("$workflowRunId: String", "$workflowRunId: UUID") // lets make an actual UUID
+    .replace(/query_fedConsensusGenomes_items/g, "ConsensusGenome");
```
ehoops-cz, line 109: why does this one have `_items` and the others don't?

ehoops-cz (approved):

resolvers.ts
```diff
@@ -945,11 +938,24 @@ export const resolvers: Resolvers = {
       // TODO (nina): call nextgen in addition to rails to get CG count
     },
     ZipLink: async (root, args, context, info) => {
-      const res = await getFullResponse(
-        `/workflow_runs/${args.workflowRunId}/zip_link.json`,
+      // const nextGenEnabled = await shouldReadFromNextGen(context);
```
bchu1, line 941: do we still need to keep this comment?

tests/UnifiedSchema.test.ts
```diff
@@ -1,10 +1,14 @@
...
+import { GraphQLSchema, lexicographicSortSchema } from "graphql";
+import { printSchemaWithDirectives } from "@graphql-tools/utils";
 import { getMeshInstance } from "./utils/MeshInstance";
 
 describe('CZ ID graphQL federation generated schema', () => {
   it('should generate a valid schema', async () => {
     const { schema } = await getMeshInstance();
 
-    expect(printSchemaWithDirectives(schema)).toMatchSnapshot();
+    // Sort the schema
```
bchu1, line 9: Nice!

utils/httpUtils.ts
bchu1, line 81: lol do you mind aadding a comment as to why this is?

utils/httpUtils.ts
```diff
@@ -72,3 +71,114 @@ export const notFound = (message: string) => {
     message: message,
   });
 };
+
+export const getFeatureFlagsFromRequest = context => {
+  return context.request.headers.get("x-should-read-from-nextgen");
```
bchu1, line 76: is this async? if so do you mind adding a return type?

bchu1: oh nvm I see it's from the fetch

utils/httpUtils.ts
```diff
@@ -72,3 +71,114 @@ export const notFound = (message: string) => {
...
+    // if the header is set, return the value
+    return true;
+  }
+  return false;
+};
+
+export const formatFedQueryForNextGen = (query: string) => {
+  let cleanQuery = query;
+  // remove fed Prefix
+  const firstCurlyBracket = query.indexOf("{");
```
bchu1, line 91: instead of this string manipulation, is it possible to just `JSON.parse()` the query and restructure it with actual objects?

utils/httpUtils.ts
```diff
@@ -72,3 +71,114 @@ export const notFound = (message: string) => {
...
+      "X-CSRF-Token": args?.input?.authenticityToken,
+      Authorization: `Bearer ${enrichedToken}`,
+    },
+    body: JSON.stringify({
+      query: formattedQuery,
+      variables: context.params.variables,
+    }),
+  });
+  if (fullResponse === true) {
+    console.log("full response", response);
```
bchu1, line 144: these console logs might be redundant with some of the other ones?

utils/httpUtils.ts
rainandbare: I actually have no idea @saq7 do you have context on this?

utils/httpUtils.ts
saq7: yeah, this particular function is not async

utils/httpUtils.ts
rainandbare: I was thinking about this but didn't go down the path because i didn't think the graphql could be `JSON.parse()` since it is not technically Javascript.

utils/httpUtils.ts
rainandbare: All these console logs we mostly for debugging. Would it be best to remove them now?

utils/httpUtils.ts
saq7: The `if` statment? That was to cover boolean conversions, etc. Basically a hedge at the time against whatever the frontend sent

resolvers.ts
rainandbare: I am keeping it here for now as an example for other devs of how to use the custom query field. I will remove it when I edit Ziplink. Does that sound okay?

tests/FormatNextGenQuery.test.ts
rainandbare: Yeah thats right, all fed queries don't perfectly match what next gen expects so we have to do some adjusting - this is for sure a brittle way to write the code.

utils/httpUtils.ts
bchu1: oh yeah, is this consensus genome stuff a leftover hack from the demo?

utils/httpUtils.ts
bchu1: maybe best to delete? I don't care that much though, if you don't feel like it just leave it :P

bchu1 (approved):

utils/httpUtils.ts
rainandbare: This is to make the Fragment part of the query work with next gen. The type that the info is on in Next Gen is different than the type that it is on in Fed server. It is a hack from the demo but I have no plans on getting rid of it.

rainandbare: This can evolve if it needs to if we end up using this function regularly. Yeah?

utils/httpUtils.ts
```diff
@@ -72,3 +71,114 @@ export const notFound = (message: string) => {
...
+export const shouldReadFromNextGen = async context => {
+  let shouldReadFromNextGen = getFeatureFlagsFromRequest(context);
+  if (shouldReadFromNextGen === true || shouldReadFromNextGen === "true" || shouldReadFromNextGen === "True") {
+    // if the header is set, return the value
+    return true;
+  }
+  return false;
+};
+
+export const formatFedQueryForNextGen = (query: string) => {
```
jfoo1984, line 88: any chance we could add a brief comment above about what this function is doing?  it's a bit hard to figure that out from reading the code.

rainandbare: good idea. It is a very questionable little piece of code.

utils/httpUtils.ts
```diff
@@ -72,3 +71,112 @@ export const notFound = (message: string) => {
...
+  if (fullResponse === true) {
+    return response;
+  } else {
+    return await response.json();
+  }
+};
+
+const checkForLogin = (responseUrl: string | null) => {
+  if (responseUrl?.includes("/auth0/refresh_token?mode=login")) {
+    throw new Error("You must be logged in to perform this action.");
```
jfoo1984, line 180: I think in the past, if you throw an error and it isn't caught, it will kill the mesh process this code was running in (although it automatically attempts to restart it).  Looking above, I believe this error should be caught and we reject the Promise, right?  Just wanted to verify so we don't kill the server processes 😅

jfoo1984 (approved):
Thanks for getting all this together in a state where we can merge it!

utils/httpUtils.ts
rainandbare: Mmm yeah, this function has been there for a while but now that the functions are modular maybe I need try catch statements in the fetchFromNextGen and the getFromRails incase they are called on their own

PR #142, opened 2024-03-03
feat: schema edits for workflows, fedWorkflows, BulkDownloadModal and fedConsensusGenomes

rainandbare:
# Pull Request

## JIRA Ticket

[CZID-9381](https://czi-tech.atlassian.net/browse/CZID-9381) & [CZID-8834](https://czi-tech.atlassian.net/browse/CZID-8834)

## Description

- switch id for producingRunId in fedConsensusGenomes (should only effect Sample Report call not Discovery View)
- add workflowRunIdsStrings to both Create Bulk Downloads and Create Bulk Downloads CSV
- add back `.toString()` for fedWorkflow runs (not used yet) in preparation for re-merging[ CZID-8834 ](https://github.com/chanzuckerberg/czid-web-private/pull/4326)

## Notes

- There are no changes here that should effect live Fed Queries! These edits are in preparation for Web app changes down the road.

## Tests

- Smoke tested Sample Report, Bulk Download Modal with CZID Web main branch in preparation for staging. 

[CZID-9381]: https://czi-tech.atlassian.net/browse/CZID-9381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CZID-8834]: https://czi-tech.atlassian.net/browse/CZID-8834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ehoops-cz (approved):

resolvers.ts
```diff
@@ -831,7 +834,7 @@ export const resolvers: Resolvers = {
           context,
         });
         return workflowRuns.map(run => ({
-          id: run.id,
+          id: run.id.toString(),
```
bchu1, line 837: Does this change need to be made to `workflowRuns` as well? (I guess this would only apply during the period between the Fed deploy and webapp deploy)

bchu1: Or if this is only going to get merged after the staging cut, doesn't matter

bchu1 (approved):

resolvers.ts
ehoops-cz: I think we might not want this change for now because the frontend changes won't go in until after this is deployed.  

Right now, BulkDownloadModal takes what this returns and then sends it to `BulkDownloadModalMutation` -> workflowRunIds ([Int]) , so wouldn't changing this to a string be a problem?

ehoops-cz (changes requested):
Question regarding the .toString() portion

ehoops-cz (approved):

PR #144, opened 2024-03-04
chore: Delete tempWorkflowRuns

bchu1:
# Pull Request

Rails will have to start hitting `workflowRuns` instead of `tempWorkflowRuns`.

Also move the .prettierrc to the top level directory instead of /json-schemas (though I still get some formatting diffs..).

## Tests

Not used in prod.


jfoo1984:
FYI, we've noted in chanzuckerberg/czid-web-private#4296 where [this reference needs to be updated](https://github.com/chanzuckerberg/czid-web-private/pull/4296/files#r1509461285).

bchu1:
> lgtm, thanks for doing this!
> 
> Do we need to coordinate the merging / release of this PR with any FE work? Or is this field not being used at all by the FE in prod yet?

FE will start using `fedWorkflowRuns` in staging as of today. I'll merge this after the staging cut so it won't affect FE

PR #148, opened 2024-03-04
feat: federate SampleForReport object

ehoops-cz:
# Pull Request

## JIRA Ticket

[CZID-9440](https://czi-tech.atlassian.net/browse/CZID-9440)

## Description

This adds federation for the SampleForReport query. 

There is logic here to pull from Rails or to acknowledge the feature flag and pull from next gen. 

### To Do In Next PR
  - deprecated - this is not where we expect it to be in the sandbox data - I will investigate and fix in a different PR
  - add clause to workflowQuery to specify "consensus-genome" rather than "bulk-download"
  - more advanced error handling if one of the queries do not work

## Tests

- Unit tests added
- Smoke Tests on the front end included testing the UI in sandbox and on local for functionality with and without the feature flag on. 

[CZID-9440]: https://czi-tech.atlassian.net/browse/CZID-9440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.meshrc.yaml
```diff
@@ -96,6 +96,18 @@ sources:
             requestSample: { pipelineVersion: "8.0" }
             responseSchema: ./json-schemas/sampleMetadata.json
             responseTypeName: SampleMetadata
+          - type: Query
+            field: SampleForReport
+            method: GET
+            path: /samples/{args.railsSampleId}.json
+            responseSchema: ./json-schemas/sampleForReportResponse.json
+            responseTypeName: SampleForReport
+          - type: Query
```
bchu1, line 105: Why are there 2?

json-schemas/sampleForReportRequest.json
bchu1, line 1: I don't think this is valid JSON schema (need the `"type"` thing)

resolvers.ts
```diff
@@ -455,6 +460,211 @@ export const resolvers: Resolvers = {
...
+      const nextGenSampleId = entitiesResp?.data.samples[0].id;
+      const workflowsQuery = `
+          query WorkflowsQuery {
+            workflowRuns(where: {entityInputs: {inputEntityId: {_eq: "${nextGenSampleId}"}}}) {
+              id
+              _id
+              railsWorkflowRunId
+              status
+              ownerUserId
+              deprecated
```
bchu1, line 594: I think this is now `deprecatedById` or something

resolvers.ts
```diff
@@ -455,6 +460,211 @@ export const resolvers: Resolvers = {
...
+      // ✅ 🤷‍♀️ input_error
+      // 🤷‍♀️ run_finalized
+      // ✅ 🤷‍♀️ wdl_version
+
+      // 2. ⏰ parse ref_fasta from entitiesResp
+      // 3. ✅ 🤷‍♀️ specify "consensus-genome" workflow type from workflowsQuery
+      // 4. 🤷‍♀️ add clause to workflowQuery to specify "consensus-genome" rather than "bulk-download"
+      // 5. ✅ parse creationSource from rawInputsJson
+
+      const entitiesQuery = `
```
bchu1, line 533: This craziness makes sense to me. Sad that we have to do it but I don't see any other way 😢

.meshrc.yaml
rainandbare: There shouldn't be!

bchu1 (approved):

PR #152, opened 2024-03-07
feat: Join NextGen data with Rails data in resolver

bchu1:
# Pull Request

## JIRA Ticket

The queries are synchronous because `workflowRuns` doesn't give us `railsSampleId`s.

## Tests

Being flag

### DIFF
--- a/example-queries/sequencing-reads-query.graphql
+++ b/example-queries/sequencing-reads-query.graphql
@@ -12,12 +12,19 @@ query TestQuery {
         }
     }) {
         id
+        technology
         taxon {
             name
         }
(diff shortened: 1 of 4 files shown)


jfoo1984 (approved):
This looks good, but to clarify your PR note, these calls need to be made synchronosly because we need the `railsSampleId` from nextGen to make the right call to the Rails API, right?

bchu1:
> This looks good, but to clarify your PR note, these calls need to be made synchronosly because we need the `railsSampleId` from nextGen to make the right call to the Rails API, right?

Yep

PR #159, opened 2024-03-08
feat: query aggregates from NextGen and reformat response

ninabernick:
# Pull Request

## JIRA Ticket

[CZID-9419]

## Description

Query CG runs from next gen and make the response look next-gen-y

## Notes

- writing endpoint so that we can either make one aggregate query per workflow (e.g. CG, AMR, mNGS each hit the aggregate endpoint) or we just do one big query by joining all the workflow run ids together. This way we can change the frontend without changing the federation layer.
- it's sort of convoluted to process the rails data into the `groupBy` format and then unprocess it into a usable format on the frontend but it fits into how the response will ultimately look on the FE

## Tests

Hit endpoint from GraphiQL and from frontend code successfully. Ran NextGen query in staging and confirmed it returned correct results.

[CZID-9306]: https://czi-tech.atlassian.net/browse/CZID-9306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CZID-9419]: https://czi-tech.atlassian.net/browse/CZID-9419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/resolvers.ts
+++ b/resolvers.ts
@@ -6,7 +6,6 @@ import {
   query_fedSamples_items,
   query_fedSequencingReads_items,
   query_fedWorkflowRunsAggregate_aggregate_items,
-  query_fedWorkflowRunsAggregate_items,
   query_fedWorkflowRuns_items,
 } from "./.mesh";
 import { processWorkflowsAggregateResponse } from "./utils/aggregateUtils";
@@ -840,12 +839,12 @@ export const resolvers: Resolvers = {
           await fetchFromNextGen({
             customQuery: convertSequencingReadsQuery(context.params.query),
             customVariables: {
-              where: input.where,
+              where: input?.where,
               // TODO: Migrate to array orderBy.
               orderBy:
-                (input.orderBy != null ? [input.orderBy] : undefined) ??
+                (input?.orderBy != null ? [input.orderBy] : undefined) ??
                 input.orderByArray,
-              limitOffset: input.limitOffset,
+              limitOffset: input?.limitOffset,
               producingRunIds: input.where?.id?._in,
             },
             serviceType: "entities",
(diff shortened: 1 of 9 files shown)


resolvers.ts
```diff
@@ -825,9 +826,9 @@ export const resolvers: Resolvers = {
           await fetchFromNextGen({
             customQuery: convertSequencingReadsQuery(context.params.query),
             customVariables: {
-              where: input.where,
-              orderBy: input.orderBy != null ? [input.orderBy] : [], // TODO: Migrate to array orderBy.
-              limitOffset: input.limitOffset,
+              where: input?.where,
```
bchu1, line 829: I'd prefer the function fail if `input` is nullish than send the wrong query to NextGen

resolvers.ts
```diff
@@ -1139,8 +1140,8 @@ export const resolvers: Resolvers = {
         const response = await fetchFromNextGen({
           customQuery: convertWorkflowRunsQuery(context.params.query),
           customVariables: {
-            where: input.where,
-            orderBy: input.orderBy != null ? [input.orderBy] : [], // TODO: Migrate to array orderBy.
+            where: input?.where,
```
bchu1, line 1143: Same here

resolvers.ts
```diff
@@ -1232,19 +1233,42 @@ export const resolvers: Resolvers = {
         context,
       });
 
-      if (!projects?.length) {
-        return [];
+      const nextGenEnabled = await shouldReadFromNextGen(context);
+      const customQuery =
```
bchu1, line 1237: nit - maybe put this in the `nextGenEnabled` block?

bchu1: Oh I see you're not returning there. Nvm

resolvers.ts
bchu1: Actually I'm not even sure how this is compiling without the `?`...ok in that case I'm fine with it lol

bchu1 (approved):

resolvers.ts
bchu1: And I'm using `?` in other places already so I don't know what I'm talking about lol

bchu1: Since you pulled it's checking `== null` above now so this isn't needed

bchu1 (approved):

PR #160, opened 2024-03-08
fix: Sample View header & pipeline version

rainandbare:
# Pull Request

## JIRA Ticket

no ticket

## Description

Make default pipeline id into a string so that it can be matched on the front end. 

## Notes

This bug was keeping the header buttons from showing on Sample Report.

## Tests

Load Sample Report with next gen flag off and see no buttons on pipeline tabs.

### DIFF
--- a/tests/__snapshots__/UnifiedSchema.test.ts.snap
+++ b/tests/__snapshots__/UnifiedSchema.test.ts.snap
@@ -1915,7 +1915,7 @@ type Sample implements EntityInterface & Node {
   alignmentConfigName: String
   basespaceAccessToken: String
   collectionId: Int!
-  createdAt: ISO8601DateTime
+  createdAt: DateTime!  
   dagVars: String
   defaultBackgroundId: Int
   defaultPipelineRunId: Int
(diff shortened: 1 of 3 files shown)


tests/__snapshots__/UnifiedSchema.test.ts.snap
```diff
@@ -2971,7 +2971,7 @@ type WorkflowRun implements EntityInterface & Node {
   _id: GlobalID!
   cachedResults: String
   collectionId: Int!
-  createdAt: DateTime!
```
rainandbare, line 2974: Do you know why my DateTime's are always trying to change? @jfoo1984

bchu1 (approved):

tests/__snapshots__/UnifiedSchema.test.ts.snap
jfoo1984: I don't have any idea, but I did some searching and found a couple links that could help, though I don't think they're our exact issue.  I'm not sure exactly where this type is coming from, but the timestamps in our sample JSONs are ISO8601 dates, so that's probably why it was changing it to that type?  I'm not really sure how to fix this from the links below though. 


https://github.com/github-community-projects/graphql-client/issues/17
https://the-guild.dev/graphql/scalars/docs/scalars/date-time

jfoo1984: is there a problem if you use `ISO8601DateTime`?

jfoo1984: might also be worth checking with The Guild team.

bchu1: Yeah I also noticed the snapshot test seems to be changing the types back and forth nondeterministically? (not completely sure though, didn't dig)

PR #166, opened 2024-03-11
fix: Add sample uploader name to joined data from Rails + Use new array orderBy argument for NextGen

bchu1:
# Pull Request

FE is now using `orderByArray` so that's what we should be sending to NextGen.

In prod today, we technically show the name of the workflow runner, and if that doesn't exist we fall back to the name of the sample uploader. Showing that in NextGen would require us to add a join to Rails in `fedWorkflowRuns`, which may not be worth doing right now given the effort and additional latency that would incur. I've created a Jira to track it for now: https://czi-tech.atlassian.net/browse/CZID-9467

Also add a tsconfig.json to Fed because apparently we didn't have one (and as a result of this, now VSCode is properly saying some variables are possibly undefined, so fix those)!

## Tests

A name should appear in the "Sample" column in the CG table in NextGen.
CG runs should be default sorted by time in NextGen.

### DIFF
--- a/example-queries/workflow-runs-query-order-by-array.graphql
+++ b/example-queries/workflow-runs-query-order-by-array.graphql
@@ -1,0 +1,21 @@
+query TestQuery {
+    fedWorkflowRuns(input: {
+        orderByArray: [{
+            startedAt: "asc"
+        }],
+        todoRemove: {
+            orderBy: "createdAt"
+            orderDir: "ASC"
+        }
+    }) {
+        id
+        entityInputs {
+            edges {
+                node {
+                    fieldName
+                    inputEntityId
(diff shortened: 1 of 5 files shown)


ehoops-cz (approved):
It looks like the frontend will continue passing orderBy until the full migration is complete and then it will swap over to orderByArray.  Is that correct?

bchu1:
> It looks like the frontend will continue passing orderBy until the full migration is complete and then it will swap over to orderByArray. Is that correct?

FE in main is already switched to `orderByArray`, but I'm keeping the support for `orderBy` in Fed for now so it's backwards compatible during the next prod push.

PR #167, opened 2024-03-11
fix: return rails data when no nextgen id available

ehoops-cz:
# Pull Request

## JIRA Ticket
none

## Description
For non-wgs workflow runs and non-migrated workflow runs, this should just return the data from rails rather than crashing.  This is expected for workflowRuns/pipelineRuns that are not WGS and for data that hasn't migrated yet.

### DIFF
--- a/resolvers.ts
+++ b/resolvers.ts
@@ -587,11 +587,18 @@ export const resolvers: Resolvers = {
       });
       console.log("entitiesResp - get 1", JSON.stringify(entitiesResp));
 
-      // Query workflows using NextGenSampleId to get in progress CG workflow runs
+      // Non-WGS workflows will not have nextGenSampleId. In this case, return sampleInfo from Rails.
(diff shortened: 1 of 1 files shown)


resolvers.ts
```diff
@@ -587,8 +587,18 @@ export const resolvers: Resolvers = {
       });
       console.log("entitiesResp - get 1", JSON.stringify(entitiesResp));
 
+      // Non-WGS workflows will not have nextGenSampleId. In this case, return sampleInfo from Rails.
+      const nextGenSampleId = entitiesResp?.data.samples?.[0]?.id;
+      if (!nextGenSampleId) {
+        console.log(`No NextGenSampleId found for railsSampleId: ${args.railsSampleId}`);
```
valenzuelaomar, line 593: I think we need an eslint-ignore-next-line comment up above this console log?

ehoops-cz: I don't think the fed server has a linter set up (although it probably should)

PR #169, opened 2024-03-11
fix: Update get conditional logic

ehoops-cz:
# Pull Request

## JIRA Ticket
none

## Description
Even when NextGen is enabled, there are times we still want to query rails.  For example, with SampleMetadata.  This ensures that the get call is actually intended for NextGen by checking both the flag and the query passed in.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 # Changelog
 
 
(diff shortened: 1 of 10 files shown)


bchu1:
Aren't there still places where we aren't passing in `customQuery` but still want it to go to NextGen? I thought that's what `formatFedQueryForNextGen()` was for.

ehoops-cz:
> formatFedQueryForNextGen

ya, I missed that function.  I think using the url for the conditional should work since we shouldn't be calling get to rails without a url

bchu1 (approved):

PR #171, opened 2024-03-11
fix: Use workflowRun IDs in inner consensusGenomes where, not sequencingRead IDs

bchu1:
# Pull Request

## JIRA Ticket

I was using the wrong IDs to filter `ConsensusGenome`s, therefore wasn't getting any back.

Also add support for query that only fetches `id` (will be used for filters).

Also fix test util assert function that wasn't properly handling single whitespace characters.

## Tests

Should be able to see `ConsensusGenome` fields in samples table in NextGen.

### DIFF
--- a/example-queries/sequencing-reads-query-id-fe.graphql
+++ b/example-queries/sequencing-reads-query-id-fe.graphql
@@ -1,0 +1,7 @@
+query DiscoveryViewFCSequencingReadsQuery(
+    $input: queryInput_fedSequencingReads_input_Input
+) {
+    fedSequencingReads(input: $input) {
+        id
+    }
+}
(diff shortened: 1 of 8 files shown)


example-queries/sequencing-reads-query-id-fe.graphql
```diff
@@ -0,0 +1,7 @@
+query DiscoveryViewFCSequencingReadsQuery(
+    $input: queryInput_fedSequencingReads_input_Input
+) {
+    fedSequencingReads(input: $input) {
+        id
+    }
+}
```
ehoops-cz, line 7: Do we care about ending files with a newline?

ehoops-cz (approved):

example-queries/sequencing-reads-query-id-fe.graphql
bchu1: For these tests it should be ok, we're stripping all whitespace anyways

PR #175, opened 2024-03-12
fix: Tsconfig to allow deployment

ehoops-cz:
# Pull Request

## JIRA Ticket
none

## Description
Something in the tsconfig was causing containers to crash during deployment.

## Notes

## Tests

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,11 +2,7 @@
   "compilerOptions": {
     "allowJs": true,
     "declaration": false,
-    "lib": [
-      "esnext",
-    ],
-    "module": "esnext",
-    "moduleResolution": "node",
+    "module": "nodenext",
     "noImplicitAny": false,
     "noImplicitReturns": false,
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -1,28 +1,18 @@
...
-    "allowSyntheticDefaultImports": true,
     "declaration": false,
-    "jsx": "react",
-    "lib": [
-      "esnext",
-      "dom"
-    ],
-    "module": "esnext",
-    "moduleResolution": "node",
+    "module": "nodenext",
```
ehoops-cz, line 5: I'm pretty sure this is the thing that actually fixed it. https://www.typescriptlang.org/tsconfig#module

ehoops-cz:
@bchu1 let me know if I should try putting back any of the other fields.  I think this will keep the type checking, but it would be helpful if you could verify (or let me know what to check)

bchu1 (approved):

PR #207, opened 2024-03-18
fix: Fix issues related to Rails calls made by fedSequencingReads when fetching IDs

bchu1:
# Pull Request

Fix 1: The list of IDs we were sending to Rails when NextGen is on was potentially way too large. See if the query will work now (if this works, then an improvement I can make after this PR is to make the 2 queries (to NextGen and Rails) parallel now that we're not sending `sampleIds` into the Rails call).

Fix 2: The Rails call when NextGen is off was grabbing the wrong IDs (it was getting workflow run IDs instead of sample IDs).

## Tests

Filters might start working now for NextGen.
Filters for NextGen off should be working again.


jfoo1984 (approved):
looks good, thanks for the explanation and fixes!

For fix 1, when you say the list of ids was potentially too large, is that in the sense that some of those IDs should've been filtered out before querying Rails, or because a large list causes performance or other issues in Rails?

bchu1:
> looks good, thanks for the explanation and fixes!
> 
> For fix 1, when you say the list of ids was potentially too large, is that in the sense that some of those IDs should've been filtered out before querying Rails, or because a large list causes performance or other issues in Rails?

I actually don't know exactly why it's broken, or if that's even the cause necessarily. Fed was spitting out an error with no message, and I wasn't getting any logs anywhere. All I know is after the Rails call the function stopped executing because logs stopped being printed. My only guess is that the query string was too long. 🤷

PR #218, opened 2024-03-19
feat: Allow security token to be passed for nextGen calls

jfoo1984:
# Pull Request

## JIRA Ticket
N/A

## Description
Support optional passing of security token to `get`/`fetchFromNextGen`.  This allows resolvers or other callers to retrieve and cache a security token across multiple calls to avoid fetching multiple security tokens in rapid succession.

## Notes
N/A

## Tests
- Unit tests added
- Deployed to sandbox and checked that discovery view and a project view loaded without error

### DIFF
--- a/utils/httpUtils.ts
+++ b/utils/httpUtils.ts
@@ -142,8 +142,6 @@ export const fetchFromNextGen = async ({
       }),
     });
 
-    checkForLogin(response?.url);
-
     if (fullResponse === true) {
       return response;
     } else {
--- a/utils/httpUtils.ts
+++ b/utils/httpUtils.ts
@@ -9,6 +9,7 @@ export const get = async ({
   serviceType,
   fullResponse,
   customQuery,
+  securityToken,
 }: {
   url?: string;
   args: any;
@@ -16,6 +17,7 @@ export const get = async ({
   serviceType?: "workflows" | "entities";
   fullResponse?: boolean;
   customQuery?: string;
+  securityToken?: "string";
 }) => {
   try {
     const nextGenEnabled = await shouldReadFromNextGen(context);
@@ -27,6 +29,7 @@ export const get = async ({
         serviceType,
         fullResponse,
         customQuery,
+        securityToken,
       });
     } else {
       if (!url) {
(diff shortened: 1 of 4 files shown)


utils/httpUtils.ts
```diff
@@ -9,13 +9,15 @@ export const get = async ({
...
   customQuery,
+  securityToken,
 }: {
   url?: string;
   args: any;
   context: any;
   serviceType?: "workflows" | "entities";
   fullResponse?: boolean;
   customQuery?: string;
+  securityToken?: "string";
```
bchu1, line 20: `string`

bchu1:
This means the resolvers have to take care of passing around the token? I wonder if it'd be better to just automatically set the token in the `context` object or something, though that does sound kind of bad.

jfoo1984:
> This means the resolvers have to take care of passing around the token? I wonder if it'd be better to just automatically set the token in the `context` object or something, though that does sound kind of bad.

Resolvers aren't required to pass around the token (we'd still generate a token if it's not provided).  Storing in the context could be viable too.

bchu1 (approved):

utils/httpUtils.ts
```diff
@@ -95,16 +98,18 @@ export const fetchFromNextGen = async ({
...
   customVariables,
+  securityToken,
 }: {
   args;
   context;
   serviceType: "workflows" | "entities";
   fullResponse?: boolean;
   customQuery?: string;
   customVariables?: object;
+  securityToken?: "string";
```
bchu1, line 109: here too

jfoo1984:
I deployed this to sandbox, but I'm not sure how to thoroughly test it.  The riskiest change is is probably remove `node-fetch`: https://github.com/chanzuckerberg/czid-graphql-federation-server/pull/218/files#diff-8bafe0a1d8508680ef748ee038281de6eb91a32b7e85d9da476019f0a0a646b6L1

I loaded up the discovery view and clicked into a project with the read from next gen flag on, so this seem sufficient?  @bchu1 @rainandbare if you have any other specific test cases, would you be able to check on sandbox?  Otherwise, this can probably be merged.  I'm going to be out the next couple days, so feel free to merge this on my behalf if you want to start using this to reduce the number of calls for the security token.

PR #230, opened 2024-03-22
fix: account for deprecated and bulk-download runs in SampleForReport query

rainandbare:
# Pull Request

## JIRA Ticket

no ticket

## Description

Had been meaning to do this when I pushed the SampleForReport query. This will make sure we are not getting any bulk downloads or deprecated runs associated with this sample. 

## Tests

- Should be able to load sample report


resolver-functions/SampleForReport.ts
```diff
@@ -0,0 +1,213 @@
...
+      query WorkflowsQuery {
+        workflowRuns(where: {
+          workflowVersion: {workflow: {name: {_eq: "consensus-genome"}}}, 
+          entityInputs: {inputEntityId: {_eq: "${nextGenSampleId}"}}
+        }) {
+          id
+          _id
+          railsWorkflowRunId
+          status
+          deprecated
```
rainandbare, line 138: @omar this is what I am trying to add

resolver-functions/SampleForReport.ts
```diff
@@ -0,0 +1,213 @@
...
+      } 
+    `;
+  const entitiesResp = await get({
+    args,
+    context,
+    serviceType: "entities",
+    customQuery: entitiesQuery,
+  });
+
+  // Non-WGS workflows will not have nextGenSampleId. In this case, return sampleInfo from Rails.
```
ehoops-cz, line 114: Looks good for now, we may need to tweak this when we get to AMR.

ehoops-cz (approved):

PR #246, opened 2024-03-26
feat: Add sorting by consensusGenomes field functionality in Fed

bchu1:
# Pull Request

Note that this is getting rid of a lot of the previously written logic in this resolver for discovery view 🎉. We no longer need all that now that this query will not be used for actual paginated data, but only for generating the listed of filtered + sorted IDs.

Also fixes the `sequencingReads` query to NextGen not sending `orderBy`.

### DIFF
--- a/resolvers.ts
+++ b/resolvers.ts
@@ -68,18 +68,17 @@ export const resolvers: Resolvers = {
     fedBulkDownloads: fedBulkDowloadsResolver,
     BulkDownloadCGOverview: BulkDownloadsCGOverviewResolver,
     fedConsensusGenomes: async (root, args, context) => {
-      /* --------------------- Next Gen ------------------------- */
-      const nextGenEnabled = await shouldReadFromNextGen(context);
-      if (nextGenEnabled) {
-        const ret = await get({ args, context, serviceType: "entities" });
-        return ret.data.consensusGenomes;
-      }
-      /* --------------------- Rails ------------------------- */
-      console.log(args.input)
       const input = args.input;
+      const nextGenEnabled = await shouldReadFromNextGen(context);
+
+      // if there is an _eq in the response then it is a call for a single workflow run result
       if (input?.where?.producingRunId?._eq) {
-        // if there is an _eq in the response than it is a call for a single workflow run result
-        // and the rails call will be like this:
+        /* --------------------- Next Gen ------------------------- */
+        if (nextGenEnabled) {
+          const ret = await get({ args, context, serviceType: "entities" });
+          return ret.data.consensusGenomes;
+        }
+        /* --------------------- Rails ----------------------------- */
         const workflowRunId = input?.where?.producingRunId?._eq;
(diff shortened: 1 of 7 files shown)


utils/queryFormatUtils.ts
```diff
@@ -150,3 +147,18 @@ export const convertSequencingReadsQuery = (query: string): string => {
 
   return query;
 };
+
+export const convertConsensusGenomesQuery = (query: string): string => {
+  return (
+    query
+      // Replace Fed variables.
+      .replace(
+        /query [\s\S]*?{/,
```
jfoo1984, line 156: Is there a way (even if verbose) to do this without regex, or can we add an explanatory comment for anyone who comes across this?  Also, is this worth adding a test for (or is it implicitly tested by other tests)?

utils/queryFormatUtils.ts
```diff
@@ -53,15 +53,12 @@ export const convertValidateConsensusGenomeQuery = (query: string): string => {
   return (
     query
       // Replace Fed variables.
-      .replace(
-        /query [\s\S]*?{/,
-        "query ($where: WorkflowRunWhereClause) {",
-      )
+      .replace(/query [\s\S]*?{/, "query ($where: WorkflowRunWhereClause) {")
```
jfoo1984, line 56: same comment about regex, could we do it another way to add a comment explaining regex?

jfoo1984 (approved):
Few minor comments, but looks good overall!

resolvers.ts
```diff
@@ -67,19 +68,22 @@ export const resolvers: Resolvers = {
...
-      }
-      /* --------------------- Rails ------------------------- */
       const input = args.input;
-      if (input?.where?.producingRunId?._eq) {
-        // if there is an _eq in the response than it is a call for a single workflow run result
-        // and the rails call will be like this:
-        const workflowRunId = input?.where?.producingRunId?._eq;
+      if (input == null) {
+        throw new Error("fedConsensusGenomes input was nullish");
+      }
```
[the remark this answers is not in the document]
bchu1: I prefer `==` because `!` is looser (interprets `0` and `""` as `false`).

utils/queryFormatUtils.ts
bchu1: It's tested by the unit tests, I'll add a comment!

resolvers.ts
```diff
@@ -163,97 +171,20 @@ export const resolvers: Resolvers = {
...
-            limit: input?.limit,
-            offset: input?.offset,
+            limit: TEN_MILLION,
+            offset: 0,
             listAllIds: false,
           }),
         args,
         context,
       });
-      if (!workflow_runs?.length) {
```
valenzuelaomar, line 179: @bchu1 removing the big blob below broke API testing :( I was using these fields to compare consensus genomes. Any way we can add them back?

PR #250, opened 2024-03-27
feat: Add includes metadata to BulkDownloadCGOverview

rainandbare:
# Pull Request

## JIRA Ticket

[CZID-9386](https://czi-tech.atlassian.net/browse/CZID-9386)

## Description

This PR connects the rails server to the federation server and concats the sample metadata from the rails server to the NextGen generated CG overview data.

## Tests

- Should be able use bulk download type Consensus Genome Overview with the NextGen flag on

[CZID-9386]: https://czi-tech.atlassian.net/browse/CZID-9386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/utils/httpUtils.ts
+++ b/utils/httpUtils.ts
@@ -30,8 +30,12 @@ export const get = async ({
       });
     } else {
       if (!url) {
-        console.error(`You must pass a url to call rails. If you meant to call NextGen, set the serviceType. url: ${url}, serviceType: ${serviceType}`);
-        throw new Error(`You must pass a url to call rails. If you meant to call NextGen, set the serviceType. url: ${url}, serviceType: ${serviceType}`);
+        console.error(
+          `You must pass a url to call rails. If you meant to call NextGen, set the serviceType. url: ${url}, serviceType: ${serviceType}`,
+        );
+        throw new Error(
+          `You must pass a url to call rails. If you meant to call NextGen, set the serviceType. url: ${url}, serviceType: ${serviceType}`,
+        );
       }
       return getFromRails({ url, args, context, fullResponse });
     }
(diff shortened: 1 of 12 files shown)


bchu1 (approved):

rainandbare:
> What was the reasoning again for having this endpoint be a custom one rather than being like a federated NextGen field? Was it because of the conflicts?

This endpoint it returns an CSV formatted array which no NextGen endpoint would do natively. We also have to go to the rails server to get the sample metadata if the next gen flag is on. We wanted to keep the logic out of the front end for now. It will make sense to put it in the front end once there is no more dual rights and once the sample metadata comes from next gen as well. But it would have to be rewritten wherever the logic was so it was just as good to put this logic on the fed server. 

What do you think?

bchu1:
> > What was the reasoning again for having this endpoint be a custom one rather than being like a federated NextGen field? Was it because of the conflicts?
> 
> This endpoint it returns an CSV formatted array which no NextGen endpoint would do natively. We also have to go to the rails server to get the sample metadata if the next gen flag is on. We wanted to keep the logic out of the front end for now. It will make sense to put it in the front end once there is no more dual rights and once the sample metadata comes from next gen as well. But it would have to be rewritten wherever the logic was so it was just as good to put this logic on the fed server.
> 
> What do you think?

Got it. Maybe the long term approach should be that NextGen gives us an actual API that returns the CSV data, instead of doing it on the FE.

utils/httpUtils.ts
```diff
@@ -62,6 +66,7 @@ export const postWithCSRF = async ({
       body: JSON.stringify(body),
     });
     checkForLogin(response?.url);
+    console.log(response);
```
jfoo1984, line 69: is this needed for debugging?

PR #294, opened 2024-04-02
fix: Use new Bulk Downloads Endpoint

rainandbare:
# Pull Request

## JIRA Ticket

no ticket 

## Description

Consume new merged bulk download rails endpoint that has all the data in one call so that we don't have to make multliple calls to the details endpoint. 

Final fix for undefined bug. 

## Notes

Dependant on this PR https://github.com/chanzuckerberg/czid-web-private/pull/4478

## Tests

- Should be able to load `/bulk_downloads` without undefined errors

### DIFF
--- a/resolver-functions/fedBulkDownloads.ts
+++ b/resolver-functions/fedBulkDownloads.ts
@@ -27,6 +27,7 @@ export const fedBulkDowloadsResolver = async (root, args, context, info) => {
     args,
     context,
   });
+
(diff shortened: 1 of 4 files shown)


ninabernick (approved):
Looks good, if you don't need the additional details about bulk download type from `download_type_details: BulkDownloadTypesHelper.bulk_download_type(bulk_download.download_type)` that gets merged in, I can remove that from Rails

PR #297, opened 2024-04-02
fix: Add taxon level to fedSequencingReads

ehoops-cz:
# Pull Request

## JIRA Ticket
none

## Description
makes the taxon level available

## Notes
I'm not sure why the snapshot update changed some dates from DateTime to ISO8601DateTime here, but it seems ok?

## Tests
updated snapshot

### DIFF
--- a/json-schemas/sequencingReadsResponse.json
+++ b/json-schemas/sequencingReadsResponse.json
@@ -126,6 +126,9 @@
                                             "properties": {
                                                 "name": {
                                                     "type": "string"
+                                                },
+                                                "level": {
+                                                    "type": "string"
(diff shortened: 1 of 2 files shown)


bchu1 (approved):

json-schemas/sequencingReadsResponse.json
```diff
@@ -126,6 +126,9 @@
                                             "properties": {
                                                 "name": {
                                                     "type": "string"
+                                                },
+                                                "level": {
```
bchu1, line 130: wait, do you have to add this to the sequencing read taxon

PR #306, opened 2024-04-04
fix: make taxon info optional in response

ninabernick:
# Pull Request

## JIRA Ticket

This wasn't causing issues in the frontend, but it was returning an error along with the fed response for the sample view consensus genome query when it was a WGS sample without taxon id or accession (which happens if you select "unknown") in the upload.

### DIFF
--- a/resolvers.ts
+++ b/resolvers.ts
@@ -103,6 +103,8 @@ export const resolvers: Resolvers = {
           context,
         });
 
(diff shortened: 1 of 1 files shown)


bchu1:
Hmm, maybe technically what we should be doing in that case is to not set `taxon` and `accession`, since the fields that were marked required in the JSON schema are required in NextGen. Instead of just returning objects that are empty?

bchu1 (approved):

PR #310, opened 2024-04-09
fix: query taxa and accessions for info in sample report

ninabernick:
# Pull Request

## JIRA Ticket

[CZID-9522]

## Description

Query taxon and accession info directly from those tables in entities using the workflow run input so that we can have this info while the workflow run is still executing. Sorts the workflow runs returned by the endpoint in ascending `executed_at` order. Changes taxon id from UUID to the `upstreamDatabaseIdentifier`, which fixes the little count over the CG icon on the mNGS report and prevents you from creating a CG run from an accession you've already used.
Adds some required fields to the query so that we can type the frontend better. I'm not 100% on the sample fields that are required.

## Notes
Discovery view also has this problem it seems.

## Tests

Unit test that in-progress run still has taxon info

[CZID-9522]: https://czi-tech.atlassian.net/browse/CZID-9522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


resolver-functions/SampleForReport.ts
```diff
@@ -79,14 +80,6 @@ export const SampleForReportResolver = async (root, args, context) => {
...
                       }
-                      accession {
-                        accessionId
-                        accessionName
-                      }
-                      taxon {
-                        id
-                        name
-                      }
                       sequencingRead {
```
ninabernick, line 83: @rainandbare are we querying for the sequencing read technology here rather than at the top level because it's just easier to process the data that way?

resolver-functions/SampleForReport.ts
```diff
@@ -158,52 +162,111 @@ export const SampleForReportResolver = async (root, args, context) => {
...
+    const accession = accessionInfo.find(
+      accession => 
+        accession.id === workflowRun.entityInputs.edges.find(
+          input => input.node.entityType === "accession"
+        )?.node?.inputEntityId
+    );
     // If !consensusGenome this is a workflow run that is in progress
     return {
       deprecated: null,
-      executed_at: workflowRun?.createdAt,
```
ninabernick, line 173: `workflowRun` not null because we're iterating over it

resolver-functions/SampleForReport.ts
```diff
@@ -158,52 +162,111 @@ export const SampleForReportResolver = async (root, args, context) => {
     customQuery: workflowsQuery,
   });
   const consensusGenomes =
-    entitiesResp.data.samples[0].sequencingReads.edges[0].node.consensusGenomes
+    samples[0].sequencingReads.edges[0].node.consensusGenomes
       .edges;
-  const workflowsWorkflowRuns = workflowsResp?.data?.workflowRuns || [];
-  const nextGenWorkflowRuns = workflowsWorkflowRuns.map(workflowRun => {
+  const workflowsWorkflowRuns: NextGenWorkflowsTypes.WorkflowRun[] = workflowsResp?.data?.workflowRuns || [];
```
ninabernick, line 167: @rainandbare this is my solution for typing the nextgen queries!

resolver-functions/SampleForReport.ts
```diff
@@ -158,52 +162,111 @@ export const SampleForReportResolver = async (root, args, context) => {
...
+
+  // Fetch taxon info from entities based on workflow run inputs
+  const taxonEntityIds = {
+    "taxon": [] as string[],
+    "accession": [] as string[],
+  };
```
rainandbare, line 173: this could be 
```suggestion
  const taxonEntityIds : { taxon: string[], accession[] } = {
    taxon: [] ,
    accession: [] ,
  };
```

resolver-functions/SampleForReport.ts
```diff
@@ -158,52 +162,111 @@ export const SampleForReportResolver = async (root, args, context) => {
...
+        nextGenWorkflowRun.rails_workflow_run_id ===
         railsWorkflowRun.id,
     );
     if (!alreadyExists) {
       dedupedWorkflowRuns.push(railsWorkflowRun);
     }
   }
+
+  // sort workflow runs by latest first so that the frontend defaults to the most recent one
+  dedupedWorkflowRuns.sort((a, b) => {return (a.executed_at < b.executed_at) ? 1 : ((a.executed_at > b.executed_at) ? -1 : 0);});
```
rainandbare, line 269: nice

rainandbare (approved):
LGTM - let's give it a shot!

PR #319, opened 2024-04-10
fix: Make response schema for CreateBulkDownload consistent between flag on/off

bchu1:
# Pull Request

Just return the ID (not currently using any of the fields anyways) if successful.

I changed the response name so it doesn't conflict with the NextGen type.

### DIFF
--- a/resolver-functions/CreateBulkDownload/CreateBulkDownload.ts
+++ b/resolver-functions/CreateBulkDownload/CreateBulkDownload.ts
@@ -48,8 +48,8 @@ export const CreateBulkDownloadResolver = async (root, args, context, info) => {
       args,
       context,
     });
-    if (railsResponse.error_message != null) {
-      throw new Error(railsResponse.error_message);
+    if (railsResponse.error != null) {
+      throw new Error(railsResponse.error);
     }
     return {
       id: railsResponse.id.toString(),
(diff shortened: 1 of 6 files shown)


resolver-functions/CreateBulkDownload/CreateBulkDownload.ts
```diff
@@ -124,11 +131,12 @@ export const CreateBulkDownloadResolver = async (root, args, context, info) => {
...
-  });
-  return res;
+  return (
+    await fetchFromNextGen({
+      args,
+      context,
+      serviceType: "workflows",
+      customQuery: runBulkDownload,
+    })
+  ).data.runWorkflowVersion;
```
rainandbare, line 141: It is possible that there would be an error from NextGen. Then data and runWorkflowVersion won't exist.

utils/httpUtils.ts
```diff
@@ -67,7 +67,11 @@ export const postWithCSRF = async ({
       body: JSON.stringify(body),
     });
     checkForLogin(response?.url);
-    return await response.json();
+    const responseJson = await response.json();
+    console.log(
+      `Rails POST to ${url} response: ${JSON.stringify(responseJson)}`,
+    );
+    return responseJson;
```
rainandbare, line 74: helpful.

rainandbare (changes requested):

rainandbare:
This is a great approach. Are we assuming that an error would already be thrown from NextGen if it is not successful?

resolver-functions/CreateBulkDownload/CreateBulkDownload.ts
bchu1: Yeah, so as of this PR https://github.com/chanzuckerberg/czid-graphql-federation-server/pull/303 we're now automatically propagating errors from NextGen through Fed's `errors`

bchu1:
> This is a great approach. Are we assuming that an error would already be thrown from NextGen if it is not successful?

Yep, basically

rainandbare (approved):

PR #325, opened 2024-04-19
fix: Update NextGen schema with new errorLabel field and pass it through SampleForReport

bchu1:
# Pull Request

What was previously returned by NextGen as `errorMessage` should've actually been the label. Now there is an `errorLabel` field and `errorMessage` contains the actual sentence explaining the error. Pass these down to the sample report in an object to what Rails returns (and what the FE expects) today.

### DIFF
--- a/resolver-functions/SampleForReport.ts
+++ b/resolver-functions/SampleForReport.ts
@@ -1,10 +1,16 @@
-import { Accession, Sample, SampleForReport, Taxon, query_SampleForReport_workflow_runs_items } from "../.mesh";
+import {
+  Accession,
+  Sample,
+  SampleForReport,
+  Taxon,
+  query_SampleForReport_workflow_runs_items,
+} from "../.mesh";
 import { get, getFromRails, shouldReadFromNextGen } from "../utils/httpUtils";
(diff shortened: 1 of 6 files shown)


resolver-functions/SampleForReport.ts
```diff
@@ -223,68 +229,79 @@ export const SampleForReportResolver = async (root, args, context)=> {
...
+        deprecated: null,
+        executed_at: workflowRun.createdAt,
+        id: workflowRun.id,
+        input_error:
+          workflowRun.errorLabel != null || workflowRun.errorMessage != null
+            ? {
+                label: workflowRun.errorLabel,
+                message: workflowRun.errorMessage,
+              }
+            : undefined,
```
jfoo1984, line 265: is this the main change in this file?  I think there's a lot of other spacing or minor changes, so wanted to make sure that this is the primary change.

jfoo1984 (approved):
looks good to me, thanks for doing this!

resolver-functions/SampleForReport.ts
bchu1: Yep, this is the main change

It'd prolly be a good idea for us to add Prettier to the auto eslint format bot if that's possible...

PR #328, opened 2024-04-19
feat: admin entities + workflows endpoints

ninabernick:
# Pull Request

## JIRA Ticket

[CZID-9529]

## Description

Adds federated `adminSamples` and `adminWorkflowRuns` endpoints that always return nextgen data, regardless of whether the flag is enabled. Used for admin sample page. Ideally, we'd just send these right through to nextgen without a custom resolver, but we don't have authorization headers set up to be able to do that yet. 

## Tests

Tested in sandbox that queries are valid

[CZID-9529]: https://czi-tech.atlassian.net/browse/CZID-9529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.meshrc.yaml
```diff
@@ -27,6 +27,20 @@ sources:
       jsonSchema:
         endpoint: ${API_URL}/
         operations:
+          - type: Query
+            field: adminSamples
+            path: /samples/index_v2.json
```
ninabernick, line 32: the `path` field here and below is arbitrary -- we are not hitting the rails backend at all. It complains if I don't have these, though. Can we put a placeholder string there?

resolvers.ts
```diff
@@ -45,6 +49,41 @@ const TEN_MILLION = 10_000_000;
 
 export const resolvers: Resolvers = {
   Query: {
+    // TODO: both of these resolvers should be removed once we can query NG directly from the FE.
+    adminSamples: async (root, args, context: any, info) => {
```
rainandbare, line 53: Can we put this in its own folder in the `resolver-functions` folder? I have a PR coming to tidy up the rest of the resolvers in this file too.

resolvers.ts
```diff
@@ -45,6 +49,41 @@ const TEN_MILLION = 10_000_000;
...
+      const response = await fetchFromNextGen({
+        args,
+        context,
+        serviceType: "workflows",
+        customQuery: query,
+        customVariables: {
+          where: args.input?.where,
+        }
+      });
+      const workflowRuns: NextGenWorkflowsTypes.WorkflowRun[] = response.data.workflowRuns;
```
rainandbare, line 81: do you need any ? here: response.data.workflowRuns

.meshrc.yaml
rainandbare: I don't think it hurts anything to put a place holder in there. I am not sure why path is required but I do know that if you put vars directly in the path the query expects the args differently (outside of the input:{}). Any other thoughts?

rainandbare (approved):
Just a couple questions!

.meshrc.yaml
jfoo1984: could we add these as `additionalTypeDefs` instead of under the Rails source?  There's a couple of existing ones for our version stuff: https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/62cf994df7bdfa9af71c76cb6bf2dc2ec8f0c923/.meshrc.yaml#L300-L307

Then you can add a resolver for those types that calls whatever you need it to.

jfoo1984: here's the resolver for the custom additional type above:
https://github.com/chanzuckerberg/czid-graphql-federation-server/blob/62cf994df7bdfa9af71c76cb6bf2dc2ec8f0c923/resolvers.ts#L1140-L1143

ninabernick: passing on this for now as these resolvers are temporary, per offline discussion

PR #334, opened 2024-04-24
chore: reuse enriched token where applicable

rainandbare:
# Pull Request

## JIRA Ticket

no ticket

## Description

Using the `fetchFromNextGen` ability to pass along security tokens within a resolver. 

Likely to see any significant improvement we would want to reuse the token on load of DiscoveryView say, but passing the token between FE and Fed completely defeats the purpose of the security token. Open to suggestions though. 

That said - my CreateBulkDownload resolver with 4 queries did get a full second of speed improvement as seen here : https://docs.google.com/spreadsheets/d/1-49awU1QQ3mB0RohDlC2HtYcghEx1Og9AmTO9Q5_1vE/edit#gid=0

## Tests

- Be able to Create bulk download - Consensus Genome (Single and Concatenated), Intermediate File Outputs
- View Discovery page
- View BulkDownloads page

### DIFF
--- a/resolver-functions/fedWorkflowRuns/fedWorkflowRuns.ts
+++ b/resolver-functions/fedWorkflowRuns/fedWorkflowRuns.ts
@@ -1,6 +1,5 @@
 import { query_fedWorkflowRuns_items } from "../../.mesh";
 import { TEN_MILLION } from "../../utils/constants";
-import { getEnrichedToken } from "../../utils/enrichToken";
 import {
   fetchFromNextGen,
   get,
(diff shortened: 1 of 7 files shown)


resolver-functions/fedWorkflowRuns/fedWorkflowRuns.ts
bchu1, line 1: i think same is true with this file

bchu1 (approved):

bchu1:
Awesome performance improvement!

Now I wonder if this would also help for any requests where make call to Rails and then a call to NextGen later...though I'm not sure we have any of those...

PR #340, opened 2024-04-30
fix: Create Bulk Download backward compatibility

rainandbare:
# Pull Request

## JIRA Ticket

no ticket

## Description

Create backward compatibility

## Notes

Sister PR is here: https://github.com/chanzuckerberg/czid-web-private/pull/4621

## Tests

- Should be able to download Consensus Genome .fa Single or Concatenated, and Consensus Genome Intermediate Output Files .zip BOTH with a prod tag or a main tag in web app


resolver-functions/CreateBulkDownload/CreateBulkDownload.ts
```diff
@@ -43,106 +43,12 @@ export const CreateBulkDownloadResolver = async (root, args, context, info) => {
...
-    };
-  }
-
-  /* --------------------- Next Gen --------------------- */
-  // get the default bulk download workflow version id from the workflow service
-  if (!workflowRunIdsStrings || workflowRunIdsStrings.length === 0) {
-    throw new Error(
-      "No Next Gen Workflow Ids provided for bulk download creation",
-    );
+    return res;
```
bchu1, line 52: I think you'll have to change the json schema to make `id` not `required`.

could also just keep the `error` check and just `return;`?

bchu1 (approved):

resolver-functions/CreateBulkDownload/CreateBulkDownload.ts
rainandbare: Yeah - that sounds good. Thanks!

